PR #1, opened 2019-02-03
Fix ODBC-74: take into account max row count in each cases

serval2412:


irodushka:
Will be merged separately

PR #209, opened 2023-11-24
Feature OOAPI

irodushka:
Here is an ODBC driver edition, adopted to OOAPI calls(https://www.ibase.ru/files/firebird/fbapi.html) instead of isc_ calls.
Since OOAPI is supported by the FB3.0 and later client library, this driver version should be used with FB3.0+ only.
This code has not yet been fully tested and is not guaranteed to work correctly in all cases.


mariuz:
@irodushka please remove the install files and install an antivirus/antimalware 

https://github.com/FirebirdSQL/firebird-odbc-driver/issues/212

https://www.bitdefender.ro/solutions/free.html

irodushka:
> @irodushka please remove the install files and install an antivirus/antimalware
> 
> #212
> 
> https://www.bitdefender.ro/solutions/free.html

Please take a look at my answer here -> https://github.com/FirebirdSQL/firebird-odbc-driver/issues/212#issuecomment-1868945644
Thanks for the authoritative advices.
.

Issue #214, opened 2023-12-29
DECFLOAT not support


Zeki-Gursoy:
I get "not yet implemented" error. I think ODBC driver 3 can't support DECFLOAT data type. When support it?

irodushka:
@Zeki
You're quite right, it doesn't. In 2024 I hope, asap)
Regards and Happy NY

сб, 30 дек. 2023 г., 00:15 Zeki ***@***.***>:

irodushka:
Hi @Zeki-Gursoy , @mariuz 

You can try 3.0.0.8 build -> https://github.com/FirebirdSQL/firebird-odbc-driver/wiki

New FB4+ types are supported in compatibility mode:
1. int128-like types are treated as varchars
2. decfloat types - as double precision (with possible precision loss)
3. with timezone - as without timezone

I will think about native support in the future, but now it's the only way to get a quick win.
Regards

irodushka:
If your question is  "why the type is string", the answer is - cause int128-like types are treated as varchars.
I'm 99% sure that fmtBcd resolves to Numeric or Decimal with size > 18.

samihanhikoski:
Seems to be int128, so ignore my comment. Works nicely as varchar

irodushka:
Be sure, I greatly appreciate any  comments and wishes!

Btw I can interpret int128 as bigint (int64), but this will result in
overflow errors if the values does not fit to int64.



пн, 19 февр. 2024 г., 18:57 Sami Hanhikoski ***@***.***>:

Zeki-Gursoy:
Dear @irodushka, now works. Thank you for this ODBC driver.

`select
	cast(123 as int128) "new int128",
	cast('123.45' as DECFLOAT(34)) "new decfloat",
	cast(current_timestamp as timestamp with time zone) "new timezone"
from rdb$database`

![resim](https://github.com/FirebirdSQL/firebird-odbc-driver/assets/30706420/89b3f7d7-e26b-4854-805b-f3bc41440b88)

irodushka:
@Zeki-Gursoy , @samihanhikoski 

Can you please take a look at the 3.0.0.9 build? I've added a compatibility binds settings to the DSN setup dialog (for windows)
It looks like:
![image](https://github.com/FirebirdSQL/firebird-odbc-driver/assets/1415214/84535bd8-e02c-47d3-9217-22c37292ca04)

[...]

PR #218, opened 2024-02-17
Issue #214 : New FB4-5 types compatibility. Build 3.0.0.8

irodushka:
Attention! Since decfloats, int128, etc types are still not supported by ODBC standard, this patch does NOT implement a complete new FB types integration, but only allows to use them in compatibility mode. The solution is that ODBC driver automatically adds "set bind of ... to legacy" directives on DB attachment.
- int128-like types are binded as varchar;
- decfloats are binded as double prec (pure legacy);
- with-time-zone types are binded as without timezone (pure legacy).

### DIFF
--- a/IscDbc/Attachment.cpp
+++ b/IscDbc/Attachment.cpp
@@ -181,6 +181,40 @@ void Attachment::createDatabase(const char *dbName, Properties *properties)
 	}
 }
 
+void Attachment::setLegacyBindingsForFB4_5()
+{
+	if (!GDS || !databaseHandle || majorFb < 4) return;
+
+	ThrowStatusWrapper status(GDS->_status);
+	ITransaction* tr = NULL;
+
+	static std::vector<std::string> cmds = {
+		"set bind of int128 to varchar",
+		"set bind of decfloat to legacy",
+		"set bind of time zone to legacy"
+	};
+
+	try
+	{
+		tr = databaseHandle->startTransaction(&status, 0, NULL);
+
+		for( auto & cmd : cmds )
+			databaseHandle->execute(&status, tr, 0, cmd.c_str(), SQL_DIALECT_V6, NULL, NULL, NULL, NULL);
+
+		tr->commit(&status);
+		tr = NULL;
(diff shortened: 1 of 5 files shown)


mrotteveel:
If you want to specify it immediately on connect, it is better to use `isc_dbp_set_bind` in the DPB instead of executing `set bind` after connect.

irodushka:
Hi Mark!

You are definitely right, isc_dpb_set_bind is the preferred way, but I
don't see any documentation about it
If you can tell me how to use this dpb, I will rewrite this code in 5 min))
Usage example would be perfect. I know it's smth semicolon-separated but
cannot find the exact format.

And another question - what if I have FB3.0 and try to attach with
isc_dpb_set_bind? Will it be ok?.. I can detect the FB server version
_after_ attach, but have no way to predict it before connecting.

Regards


сб, 17 февр. 2024 г. в 17:44, Mark Rotteveel ***@***.***>:

mrotteveel:
The `isc_dpb_set_bind` option takes a semi-colon separated list of the binds you want to set up (basically what you'd otherwise put after `set bind to` in the statement. So to replace your `set bind` statements, you would use `isc_dpb_set_bind` with string value `"int128 to varchar;decfloat to legacy;time zone to legacy"`. Unless I'm mistaken, older versions of Firebird should ignore unknown DPB items.

BTW: you can find the documentation in `doc/sql.extensions/README.set_bind.md` in your Firebird 4 or Firebird 5 installation.

mrotteveel:
Or otherwise: https://github.com/FirebirdSQL/firebird/blob/master/doc/sql.extensions/README.set_bind.md

mrotteveel:
To reinforce my previous wishy-washy statement: older Firebird versions definitely ignore unknown DPB items.

irodushka:
Great!
Yes, it works as expected.
Thank you for your review!)  Just committed to the repo.

Regards

сб, 17 февр. 2024 г. в 18:50, Mark Rotteveel ***@***.***>:

PR #237, opened 2024-12-16
Unstable operation with a large number of connections

NickNevzorov:
https://github.com/FirebirdSQL/firebird-odbc-driver/issues/236

### DIFF
--- a/IscDbc/Attachment.cpp
+++ b/IscDbc/Attachment.cpp
@@ -357,7 +357,7 @@ void Attachment::openDatabase(const char *dbName, Properties *properties)
 
 	if( ( check_status.getState() & IStatus::STATE_ERRORS ) == 0 )
 	{
-		for (auto p = result; p < result + sizeof (result) && *p != isc_info_end;)
+		for (auto p = result; p < result + sizeof (result) && *p != isc_info_end && *p != isc_info_truncated;)
 		{
 			char item = *p++;
 			int length = GDS->_vax_integer (p, 2);
--- a/IscDbc/Attachment.cpp
+++ b/IscDbc/Attachment.cpp
@@ -42,11 +42,11 @@ static char databaseInfoItems [] = {
 	isc_info_db_id,
 	isc_info_db_sql_dialect,
 	isc_info_base_level,
-	isc_info_user_names,
 	isc_info_ods_version,
 	isc_info_firebird_version,
 	isc_info_version, 
 	isc_info_page_size,
+	isc_info_user_names,
 	isc_info_end 
 	};
 
--- a/WriteBuildNo.h
+++ b/WriteBuildNo.h
@@ -4,4 +4,4 @@
 // Note - there must be two tabs between BUILDNUM_VERSION and 
 // the actual number, otherwise the makefile for linux will not
 // pick up the value.
-#define BUILDNUM_VERSION		15
+#define BUILDNUM_VERSION		16


IscDbc/Attachment.cpp
```diff
@@ -357,7 +357,7 @@ void Attachment::openDatabase(const char *dbName, Properties *properties)
 
 	if( ( check_status.getState() & IStatus::STATE_ERRORS ) == 0 )
 	{
-		for (auto p = result; p < result + sizeof (result) && *p != isc_info_end;)
+		for (auto p = result; p < result + sizeof (result) && *p != isc_info_end && *p != isc_info_truncated;)
```
mrotteveel, line 360: It would be better to explicitly handle isc_info_truncated (e.g. retrying with a larger buffer, or throwing an exception; not sure what the best would be here), instead of considering it just the same as isc_info_end

NickNevzorov: @mrotteveel a larger buffer will not solve the problem. The UserName parameter is repeated in the buffer itself an infinite number of times. 

Calling the function again is an option, but even with such a minimal fix everything starts working for me under heavy load

mrotteveel: I'm guessing I'm missing some context here (like the contents of `databaseInfoItems`), but if that is the case, that sounds like a bug in Firebird.

mrotteveel: OK, I guess it's the `isc_info_user_names` info item; which wouldn't be "infinite", it just contains the user names of all current attachments to the database, so with a lot of attachments, you'll get a lot of user names. I'm not sure why it is requested though, that seems unnecessary and not useful information, and I wonder about how it's getting used in this code.

mrotteveel: As in, it seems to be used to determine the actual username of the current user, and to determine if they are an admin or not, but there is no actual guarantee that the first name returned is of the current attachment (maybe on Classic it would be). Probably it should request `fb_info_username` instead, but that was only introduced in Firebird 4.

mrotteveel: Also, with the truncation, the information items isc_info_ods_version, isc_info_firebird_version, isc_info_version, and isc_info_page_size, are never received as they come after isc_info_user_names in `databaseInfoItems`, which means the connection might be missing information.

NickNevzorov: @mrotteveel you are right. After fb_info_username  there is other data. It is also necessary to increase the buffer 

users

mrotteveel: Note that this only affects SYSDBA and users exercising RDB$ADMIN, as otherwise `isc_info_user_names` only reports the username for the current connection.

NickNevzorov: I suggest increasing the buffer to 64K
This will be enough for almost 7K simultaneous connections, one might say an unrealistic scenario:

`char result [64 * 1024];`

mrotteveel: If you have very long UTF8 user names, it might be less than that (maximum username length is 252 bytes since Firebird 4, so each info item can take up to 256 bytes).

I did some experimenting with Jaybird, and it seems that Firebird does actually report the username of the current connection first (though I'm not sure if it's guaranteed). As such it seems that keeping the buffer size as is and moving `isc_info_user_names` to the second-last item of `databaseInfoItems` (immediately before `isc_info_end`) will also work well enough with your current solution to treat `isc_info_truncated` the same as `isc_info_end`, because truncation will only be reported with a later username.

mrotteveel: @NickNevzorov Even if you put `isc_info_user_names` last (before `isc_info_end`)? I know the versions can become relatively long, but I wouldn't expect it to be that long.

mrotteveel: I think that if you use a native client, it rewrites the buffer to append its own version information to the version items, so that might be problematic indeed. I guess that maybe the `isc_info_user_names` request should be done separately (and if Firebird 4.0 or higher using `fb_info_username`.

NickNevzorov: @mrotteveel It is difficult to say what data will be transferred first in new versions of Firebird. 64K buffer will provide greater reliability, especially since this is a one-time operation, the memory is freed when exiting the procedure

mrotteveel: Looking at the Firebird source code, information items are written in the order they're requested, so putting `isc_info_user_names` last should alleviate the issue, and I would expect truncation to occur after all other information items have been retrieved and at least one `isc_info_user_names` item (which would be sufficient for the logic in this class to work).

NickNevzorov: Done

Issue #231, opened 2024-09-26
Publish build artifacts to the `Releases` section of the GitHub project.


fdcastel:
The current release process is currently being done manually (I believe) and involves editing [the Wiki homepage](https://github.com/FirebirdSQL/firebird-odbc-driver/wiki).

Ideally, each release should be published into [Releases section](https://github.com/FirebirdSQL/firebird-odbc-driver/releases) of the project. 

Why am I asking this? Because I'm also the maintainer of the [`firebird-odbc` chocolatey package](https://community.chocolatey.org/packages/firebird-odbc), and I'm trying to automate it, too 😅. 

With the build artifacts being published to GitHub Releases, we can easily retrieve them via the GitHub API. An [example](https://github.com/fdcastel/firebird-docker/blob/master/firebird-docker.build.ps1#L105) of this can be seen in the Firebird-Docker project.

fdcastel:
_Just for the sake of completion._

Today we have the following artifact listed on the Wiki homepage as the download for Windows:

https://github.com/FirebirdSQL/firebird-odbc-driver/actions/runs/11053399614/artifacts/1982531858

However, this is a `.zip` file containing both `x64` and `x86` installers.

For the chocolatey package, [we need each installer to be available separately](https://github.com/fdcastel/chocolatey-packages/blob/master/firebird-odbc/tools/chocolateyinstall.ps1#L4-L5) for download, similar to what was done on SourceForge.

irodushka:
Well...
 
I don't think making a release for every bugfix is ​​a good approach) The release is something of a milestone and should contain some significant changes.
From the other hand, I understand your problem, appreciate your work and would like to help.

The builds are not manual, I have a "git actions" for it - one for Windows and second for Linux. Every commit in master (e.g. merging a pull-request) triggers an automated build process. Yes, editing the wiki page is done manually)) but links lead to git action artifacts. Packing artifacts into a zip archive is the default behavior of the git action publishing step, maybe it can be changed with some settings, I have to investigate.

But. Maybe we should integrate our automation processes?
Theoretically, I could push something to inform your automated process about the new build, and also provide it with links to artifacts?..

fdcastel:
> I don't think making a release for every bugfix is ​​a good approach 

Agreed. It wasn't my intention to suggest the opposite. 👍🏻



> Packing artifacts into a zip archive is the default behavior of the git action publishing step, maybe it can be changed with some settings, I have to investigate.

[...]

PR #245, opened 2025-05-31
[WIP/Do not merge] New GitHub Actions workflow.

fdcastel:
Initial work to solve #231. Based on @asfernandes work on Firebird project. 🚀 

GitHub release can be seen [here](https://github.com/fdcastel/firebird-odbc-driver/releases).

## Changes

- **The `release` workflow is now triggered when a tag starting with `v` is pushed.**
- Replaced `pguyot/arm-runner-action` with native `ubuntu-22.04-arm` runners
- Fixed build paths from `Builds/Msvc2019` to `Builds/MsVc2022.win`
- Unified artifact naming: `firebird-odbc-{platform}-{arch}`
- Single workflow file instead of separate per-platform workflows

## Advantages over previous pipeline

✅ **Native ARM64 Execution**: No more Raspberry Pi emulation  
✅ **Faster Builds**: Native runners are much faster than emulated ones  
✅ **Simplified Maintenance**: Matrix strategy reduces code duplication  
✅ **Better Reliability**: No complex emulation dependencies  
✅ **Consistent Environment**: Same OS versions across architectures  
✅ **Automatic Releases**: GitHub releases created automatically with all artifacts  

## ToDo:

- [ ] Add Windows-ARM64 builds
- [ ] Refactor InnoSetup script
  - [ ] Remove MakePackage.bat and `sed`.
  - [ ] Call `iscc.exe` direcly, [using /F](https://jrsoftware.org/ishelp/index.php?topic=compilercmdline) to override `OutputBaseFilename` 
  - [ ] Normalize names (`amd64` for Linux, `x64` for Windows)
- [ ] Automatic release notes

RFC @irodushka @asfernandes @mrotteveel


.github/workflows/release.yml
```diff
@@ -0,0 +1,419 @@
+name: release
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+permissions:
+  contents: write
```
mrotteveel, line 9: I think only the create-release job needs write permissions, so it's better to set this to read, and set the write permission only in the create-release job.

.github/workflows/release.yml
```diff
@@ -0,0 +1,419 @@
...
+    - name: Create Release
+      uses: softprops/action-gh-release@v2
+      with:
+        files: |
+          firebird-odbc-windows-x86.zip
+          firebird-odbc-windows-x64.zip
+          firebird-odbc-linux-x64.tar.gz
+          firebird-odbc-linux-arm64.tar.gz
+          firebird-odbc-windows-x86-installer.exe
+          firebird-odbc-windows-x64-installer.exe
```
mrotteveel, line 417: The version should be part of the file names of the published artifacts.

mrotteveel: (I placed it here, but it starts with the file names generated in the other jobs)

.github/workflows/release.yml
```diff
@@ -0,0 +1,419 @@
...
+  push:
+    tags:
+      - 'v*'
+
+permissions:
+  contents: write
+
+jobs:
+  # Stage 1: Build Windows binaries (parallel)
+  build-windows-x86:
```
mrotteveel, line 13: I'm not 100% sure due to the artifact handling, but I guess it should be possible to use `strategy.matrix` to define the build-windows, package-windows and installer-windows jobs once, instead of individually for x86, x64 and arm64.

fdcastel: Nice catch! I'm still getting the hang of this (and the whole 'strategy matrix' concept). Looking into it now... 👍🏻

irodushka:
Great job! I will take a look next week!

сб, 31 мая 2025 г., 22:48 Mark Rotteveel ***@***.***>:

.github/workflows/release.yml
fdcastel: Done in https://github.com/FirebirdSQL/firebird-odbc-driver/pull/245/commits/37b83fb0a64fe750774f45610720ef2ea00b4252

.github/workflows/release.yml
fdcastel: Done in https://github.com/FirebirdSQL/firebird-odbc-driver/pull/245/commits/ed49e7547d621afe8386741336ded57326c465d0

fdcastel:
@irodushka Before I can make changes to `Install\Win32\OdbcJdbcSetup.iss`, I need to know the encoding it was saved with. 

The file contains some Russian translations, so I want to ensure it's correctly converted to UTF-8 without corrupting the text.

In VSCode, when opened with the "Windows 1251" encoding, it displays as:

image

fdcastel:
Can we remove these old Windows build environments?

```
Builds\MsSDK64.win
Builds\MsVc60.win
Builds\MsVc70.win
Builds\MsVc80.win
Builds\MsVc90.win
```

fdcastel:
Something seems off with the MSBuild project configuration.

- Even when invoking `msbuild` with [`/p:Configuration=Release`](https://github.com/fdcastel/firebird-odbc-driver/actions/runs/15366491611/job/43240077283#step:5:1),

- The linker is still [being executed](https://github.com/fdcastel/firebird-odbc-driver/actions/runs/15366491611/job/43240077283#step:5:468) with `/DEBUG /PDB:"x64/Release/FirebirdODBC.pdb"`.

That’s why `.pdb` files are being generated.

P.s.: The [latest release build](https://github.com/FirebirdSQL/firebird-odbc-driver/actions/runs/13812419198/job/38636908620#step:7:129) shows the same issue. Are the currently distributed binaries actually debug builds? Or am I missing something?

mrotteveel:
> Something seems off with the MSBuild project configuration.
> 
>     * Even when invoking `msbuild` with [`/p:Configuration=Release`](https://github.com/fdcastel/firebird-odbc-driver/actions/runs/15366491611/job/43240077283#step:5:1),
> 
>     * The linker is still [being executed](https://github.com/fdcastel/firebird-odbc-driver/actions/runs/15366491611/job/43240077283#step:5:468) with `/DEBUG /PDB:"x64/Release/FirebirdODBC.pdb"`.
> 
> 
> That’s why `.pdb` files are being generated.

This is caused by `<GenerateDebugInformation>true</GenerateDebugInformation>` in the definitions in `OdbcFb.vcxproj` for `Release|Win32` and `Release|x64`

fdcastel:
> What's the point of excluding pdb files from the Release build?

Sorry for the confusion, @irodushka. I reverted this in later commits.

> If we want to reduce the installation packages sizes, it's better to store pdb as a separate artefact for optional download/installation.

That's exactly what I did 😉 --  I added new packages with the `-debug` suffix. Please see them [here](https://github.com/fdcastel/firebird-odbc-driver/releases/tag/v3.0.16-temp4).

irodushka:
> > What's the point of excluding pdb files from the Release build?
> 
> Sorry for the confusion, @irodushka. I reverted this in later commits.
> 
> > If we want to reduce the installation packages sizes, it's better to store pdb as a separate artefact for optional download/installation.
> 
> That's exactly what I did 😉 -- I added new packages with the `-debug` suffix. Please see them [here](https://github.com/fdcastel/firebird-odbc-driver/releases/tag/v3.0.16-temp4).

Do I understand right that *-debug packages contain Debug build artefacts? Debug dll binary and a corresponding pdb?
What I want to say -  debug and release dlls are different, and it's not just about debug info. The Release dll contains compiler optimizations, it just _different_ binary code. So, a pdb file from a Debug build cannot be used to debug a Release dll.

irodushka:
Well... and where's my favorite VirusTotal action?..)
It's not just for fun. There are sporadic troubles with InnoSetup - Windows antivirus detects malware in Inno binaries from time to time. It's a real headache.

fdcastel:
> So the regular (wo suffix) windows package and the -debug package contain the same dll binary - the release one?

Exactly! 👍🏻 

> And the only difference between the normal & debug packages is the pdb file inside?

Exactly! 👍🏻 

> Let's use cmake style naming, it has a special build type "RelWithDebInfo" = release_with_debug_info.

Sorry. Not totally clear on what you mean 😅. But if you can show me a few examples, I’ll tailor it exactly how you want.

irodushka:
I mean to replace the package suffix _debug with _relwithdebinfo or
_release_with_debuginfo

пн, 2 июн. 2025 г., 19:15 F.D.Castel ***@***.***>:

fdcastel:
> There are sporadic troubles with InnoSetup...

InnoSetup is built with Delphi, which --  for reasons only antivirus developers seem to understand -- tends to attract these kinds of false positives like a magnet.

By the way, I'm also planning to replace InnoSetup with [WIX](https://www.firegiant.com/wixtoolset/), unless anyone has objections. WIX can generate MSI files, which are generally more suitable for enterprise environments.

irodushka:
I know)
Not an objections but a benediction))

пн, 2 июн. 2025 г., 19:24 F.D.Castel ***@***.***>:

irodushka:
> > I mean to replace the package suffix _debug with _relwithdebinfo or _release_with_debuginfo
> 
> I found the improvement a bit _subtle_, to be honest. 😄
> 
> I would lean toward following the approach [used by the Firebird Project]

Yes, it' fine for me.

fdcastel:
> For Ubuntu only. Yes, tried the ARM build on Raspberry, and it worked fine. 

🎉 Fantastic news! I'll investigate.

> Last time I looked, there were no Windows ARM runners available on github. 

Look again: 😄👀
https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/

irodushka:
I can explain briefly - no)
When I inherited this code, I also wondered - what about tests?) Nope, the
answer was.

Frankly speaking - I have some tests, it's impossible to maintain such a
... project without them, and it would be a good deal to put them in order
and commit to the repo. Hope I can do this in some future.



пн, 2 июн. 2025 г., 19:41 F.D.Castel ***@***.***>:

fdcastel:
> I would like to mention that it makes sense to retain the ability to do autobuilds for every commit into the master (not for releases only but for intermediate feature-branches merges).

Sure! I was just talking about this with @asfernandes. Take a look at [this new workflow proposal](https://github.com/FirebirdSQL/firebird-docker/pull/30#issuecomment-2934445978) I suggested for another project.



> Another interesting question - what happens when some of the builds in the strategy matrix fail? I mean - windows build is ok, linux ok, but ARM fails. Does this mean that the entire build failed?

Short answer: Yes.

The BUILD jobs are marked with [`fail-fast: false`](https://github.com/FirebirdSQL/firebird-odbc-driver/pull/245/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R16) which means the WILL PROCEED in case some parts of matrix fail.

However the RELEASE job [requires](https://github.com/FirebirdSQL/firebird-odbc-driver/pull/245/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R236-R241) a successful run of its dependent jobs. And the packages will not be published.

However, the RELEASE job [requires](https://github.com/FirebirdSQL/firebird-odbc-driver/pull/245/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R236-R241) all dependent jobs to complete successfully; otherwise, the packages will NOT be published.



> Or can we issue a warning that the build succeeded for some platforms and failed for the others?

GitHub will let us know for sure.

irodushka:
Hi again @fdcastel 

BTW I have a PR https://github.com/FirebirdSQL/firebird-odbc-driver/pull/248
It's a modern cmake for the ODBC driver build (for linux) replacing an ancient plain C-style makefile. At least smoke-tested, but definitly needs more research.
Maybe it's a good idea to merge it to your branch, what do you think about it?

fdcastel:
> It's great to see you again! Those tests... afraid they will happen in some a bit more distant futurious future)

Don’t waste time polishing them—just send them over. Any format is fine, even if they don’t work.

I’m building a brand-new test suite from the ground up. What I really need are the cases and the ideas. I’ll take care of the rest.

fdcastel:
> BTW I have a PR #248 It's a modern cmake for the ODBC driver build (for linux) replacing an ancient plain C-style makefile. 

Fantastic! I’ll take a look at #248 as well.

It came at just the right time -- the current MSBuild is giving me some headaches, too.

PR #247, opened 2025-06-07
Unicode support on Linux

aafemt:


irodushka:
Hi @aafemt 

I see this's really a huge job.
But...

I don't like so many changes in so many source files. Looks like over-complicated and resembles a patchwork quilt. Sorry.
And this is just an initial work...)

I prefer not to increase the FB ODBC driver code entropy,  since it's already very very high. 
And the lack of testing tools, as well as near-to-zero test coverage, make such a changes vastly dangerous.
And all this only to support a case that has never been seen since the FB ODBC driver foundation)) i.e. "for ages")

What options can I suggest:
1. You make a fork of the ODBC driver (concerning all this unicode stuff) and maintain it by your own, occasionally merging a master branch from here to your fork. And I can mention your fork somewhere here.
2. _OR_ Try to make your changes more systemic. For example, you can create some wrapper/helper class(es) that encapsulate all the charset conversion logic. As the first step, you implement these helpers into the existing codebase (i.e. pure refactoring, without any new functionality). We merge it and test it and it's okay. We can then override these helper layer to implement some _new_ functionality, that will resolve a ticket #247. If we do this carefully, we can switch between legacy and new unicode conversion logic simply with a #define or some feature-flag.
3. _OR_ Cover all the code you touch with unit tests, for all possible scenarios & platforms, and consider merging your patches as is. Suppose that's a kind of non-science fiction)

Frankly speaking, I prefer option (2).

Regards
Yuri

PR #248, opened 2025-07-05
Modern cmake for linux build

irodushka:


fdcastel:
For the record: this work is already covered in #275.

PR #275, opened 2026-02-10
V4 - Firebird ODBC Driver Modernization

### LINKED ISSUE #231: Publish build artifacts to the `Releases` section of the GitHub project.

### LINKED ISSUE #244: Unicode support
Currently the driver has two issues with Unicode functions:
1. They are just front-end to ANSI functions that defeats their purpose: ability to work with symbols outside of ANSI codepage.
2. A couple of bugs in ANSI<->Unicode conversion code assumes that `SQLWCHAR` is `wchar_t` and `sizeof(wchar_t)` is 2. The former makes the driver non-working with unixODBC and the latter - with iODBC.

I see following solutions for it:
1. Drop charset option and fix the connection charset to NONE, performing conversion between storage charset and ANSI/Unicode on the client side. It will cause problems with server error messages and string literals in queries effectively limiting them to ASCII.
2. Automatically set connection charset to match ANSI codepage (current locale) for `SQL*Connect` functions and fix it to UTF-8 for `SQL*ConnectW` functions. It will solve problems above but cause a new one with applications that use ANSI connection while still want to work with Unicode data using `SQLGetData`.
3. Set connection charset to UTF-8 unconditionally and perform ANSI/Unicode conversions on the client side. No problems above but if the database charset is not UTF-8, a server-side conversion is enforced causing a higher load. Besides, it requires more changes in codebase.

So, no perfect solution. Or may be anyone see one?..

### LINKED ISSUE #246: Manually update latest GitHub release to latest version
While we are working on #245... 

The [Releases section](https://github.com/FirebirdSQL/firebird-odbc-driver/releases) of the GitHub project currently has assets of an outdated version (3.0.0.11) that contains a [critical bug](https://github.com/FirebirdSQL/firebird-odbc-driver/issues/233).

Kindly update it to **version 3.0.0.16** as soon as possible, as the current state may lead to confusion and serious issues for new users.

### LINKED ISSUE #258: MSI installer
Please generate installer in MSI format.
Thanks

fdcastel:
# v4 - Firebird ODBC Driver Modernization

- Replace #248 
- Replace #247 
- Replace #245
- Fix #258 
- Fix #246
- Fix #244 
- Fix #231
- ... and many others (see [Docs/FIREBIRD_ODBC_MASTER_PLAN.md](https://github.com/fdcastel/firebird-odbc-driver/blob/v4-pr/Docs/FIREBIRD_ODBC_MASTER_PLAN.md)).

**Builds available** at https://github.com/fdcastel/firebird-odbc-driver/releases/tag/v4.0.1-preview.

## Summary

This PR is a comprehensive modernization of the Firebird ODBC driver, addressing crash bugs, ODBC specification violations, Unicode data corruption on Linux/macOS, and significant performance bottlenecks. It also replaces the build system with CMake, adds a 401-test Google Test suite, and reorganizes the source tree.

The changes are organized into 4 commits for easier review:

1. **`build: remove legacy build files, vendored headers, and dead code`** -- 383 files removed. Safe to skim; this is just deletions.
2. **`feat: modernize driver -- CMake build, source reorg, fixes and improvements`** -- The core changes. Source moved to `src/`, all driver improvements applied. Git detects 137 renames so `git diff -M` shows only the actual code changes (~5,300 lines of real modifications).
3. **`test: add comprehensive Google Test suite (401 tests, 34 suites)`** -- New tests only. Can be reviewed independently.
4. **`docs: add project documentation and developer guidelines`** -- Documentation only.

> **Reviewing tip**: For commit 2, use `git diff -M` or GitHub's "detect renames" option to see only the meaningful code changes rather than full file deletions + additions.

---

## Motivation

I started this project to address real bugs I was hitting while using the Firebird ODBC driver in production. What began as a few targeted fixes grew into a larger effort as I kept finding interconnected issues -- the crash bugs led me to the GUARD macro pattern, which led me to entry-point hardening, which revealed the missing SQLSTATE mappings, and so on.

The driver has a solid foundation -- it correctly uses the Firebird OO API for connections, transactions, statements, and result sets. But it has accumulated technical debt over its 20+ year history, and some areas (Unicode handling, error mapping, safety at the API boundary) needed significant work.

I understand this is a large PR and may take time to review. I'm happy to discuss any changes, split things differently, or revert specific decisions if they don't align with the project's direction.

This work also led to a secondary project: [ODBC Crusher](https://github.com/fdcastel/odbc-crusher). What began as a tool to investigate and learn more about ODBC drivers has since grown into a fully usable tool (still under active development).

---

### 🔴 Crash Fixes

| Issue | Fix |
|-------|-----|
| `GUARD_HDESC`/`GUARD_HSTMT`/`GUARD_HDBC` macros dereference handles **before** checking for null | Added `NULL_CHECK` macro -- check first, then dereference |
| `SQLCopyDesc` crashes when source descriptor has no records | Null guard for `sour.records` + early return when `headCount == 0` |
| `OdbcObject::postError` uses `sprintf` into 256-byte stack buffer | Changed to `snprintf` with 512-byte buffer |
| 64 unsafe C-style exception downcasts: `(SQLException&)ex` | Replaced with direct `catch (SQLException& exception)` clauses |
| No null checks at ODBC entry points -- invalid handles cause immediate access violations | Added explicit null checks; return `SQL_INVALID_HANDLE` |

### 🟡 ODBC Specification Compliance

- **SQLSTATE mapping**: Previously only 3 SQL error codes and ~19 ISC codes had explicit SQLSTATE mappings. Now: 121 SQLSTATE entries, 100+ ISC mappings, 130+ SQL code mappings, with dual ODBC 2.x/3.x strings.
- **SQLGetDiagRec**: returned `SQL_NO_DATA_FOUND` (ODBC 2.x constant) instead of `SQL_NO_DATA`.
- **SQLGetDiagField**: dereferenced `StringLengthPtr` without null check.
- **SQLGetInfo**: wrong storage size for numeric InfoTypes; crashed on `NULL InfoValuePtr`.
- **SQLCloseCursor**: returned `SQL_SUCCESS` when no cursor was open (should be SQLSTATE 24000).
- **SQLDescribeColW**: returned `SQL_CHAR`/`SQL_VARCHAR` instead of `SQL_WCHAR`/`SQL_WVARCHAR`.
- **SQLSetConnectAttr**: silently accepted unsupported attributes (no default error path).
- **SQLSetStmtAttr**: missing cursor-state validations (24000/HY011).
- **SQL_ATTR_ODBC_VERSION**: `SQLGetEnvAttr` always returned `SQL_OV_ODBC3` regardless of what was set.
- **SQL_SCHEMA_USAGE**: checked `supportsCatalogsInIndexDefinitions()` instead of the schema equivalent.
- **ODBC 3.8**: `SQL_OV_ODBC3_80` now accepted; driver reports `"03.80"`; `SQL_ATTR_RESET_CONNECTION`, `SQL_GD_OUTPUT_PARAMS`, `SQL_ASYNC_DBC_FUNCTIONS` implemented.
- **SQL_ASYNC_MODE**: was incorrectly reporting `SQL_AM_STATEMENT` while rejecting async enable with HYC00. Now correctly reports `SQL_AM_NONE`.

### 🟢 New Features

- **Savepoint isolation**: Per-statement SAVEPOINT/RELEASE/ROLLBACK TO SAVEPOINT when `autoCommit=OFF`. Failed statements don't corrupt the transaction.
- **IBatch (Firebird 4+)**: Array parameter execution now uses `IBatch` for a single server roundtrip instead of N individual roundtrips. Falls back to row-by-row for Firebird 3.x. Inline BLOB support via `BLOB_ID_ENGINE` policy.
- **SQL_GUID**: `CHAR(16) CHARACTER SET OCTETS` (FB3) and `BINARY(16)` (FB4+) are mapped to `SQL_GUID`. All conversion methods implemented.
- **Query timeout**: `SQL_ATTR_QUERY_TIMEOUT` implemented using `IAttachment::cancelOperation(fb_cancel_raise)` with a timer thread. Returns SQLSTATE `HYT00` on expiry.
- **SQLCancel**: Now functional -- calls `cancelOperation()` to interrupt in-flight operations.
- **ConnSettings**: New connection string parameter to execute SQL on connect.
- **Server version detection**: `getServerMajorVersion()`/`getServerMinorVersion()` for version-gated behavior (e.g., FB4+ types like INT128, DECFLOAT, BINARY/VARBINARY).
- **Connection pool reset**: `SQL_ATTR_RESET_CONNECTION` rolls back pending transactions, closes open cursors, and resets statement attributes.

### ⚡ Performance

- **SRWLOCK**: Replaced Win32 kernel `Mutex` (`CreateMutex`/`WaitForSingleObject`) with user-mode `SRWLOCK` on Windows. ~20ns vs ~1-2μs per lock/unlock.
- **64-row prefetch buffer**: `IscResultSet` fetches 64 rows at a time from Firebird, serving subsequent `SQLFetch` calls from the buffer. Measured: **10.88 ns/row** for 10K×10 INT columns (embedded).
- **ConvertingString stack buffer**: 512-byte stack buffer eliminates virtually all W-API heap allocations.
- **Native UTF-16 metadata caching**: `OdbcString` w-cache in `DescRecord` -- `SQLDescribeColW`, `SQLColAttributeW`, `SQLGetDescFieldW`, `SQLGetDescRecW`, `SQLGetDiagRecW`, `SQLGetDiagFieldW` all read from cached UTF-16 strings via `memcpy`, bypassing `ConvertingString` entirely.
- **LTO**: Link-Time Optimization enabled for Release builds.
- **Miscellaneous**: BLOB object pooling, `Value::getString` buffer reuse, `clearErrors` fast path, `std::to_chars` for float→string, `ODBC_FORCEINLINE` on hot helpers, `alignas(64)` on `Sqlda::buffer`.

### 🔤 Unicode

The original Unicode implementation assumed `SQLWCHAR == wchar_t`, which is only true on Windows. On Linux (`wchar_t` = 4 bytes), this caused complete data corruption for all W-API functions.

- Introduced `ODBC_SQLWCHAR` typedef (always 16-bit on all platforms).
- All codec signatures and `*ToStringW` conversion functions now use `SQLWCHAR*` directly.
- Unified to a single UTF-8↔UTF-16 codec (`Utf16Convert.h`).
- `CHARSET` defaults to `UTF8` when not specified in the connection string (was `NONE`, which used locale-dependent `mbstowcs` on Linux).
- FSS codec and `convVarStringSystemToStringW` fixed to use `SQLWCHAR*` instead of `wchar_t*`.

### 🏗️ Build System & Infrastructure

- **CMake 3.20+**: Replaces 14 platform-specific build configurations.
- **FetchContent**: Firebird public headers fetched from GitHub (pinned to v5.0.2). Google Test and Google Benchmark fetched for testing.
- **Git-based versioning**: Version extracted from tags via `cmake/GetVersionFromGit.cmake`.
- **WiX v5 MSI installer**: `installer/Product.wxs` for Windows deployment.
- **CI**: GitHub Actions for Windows x64 and Linux x64/ARM64, with Firebird 5.0 database for integration tests.
- **Source reorg**: All source moved to `src/` and `src/IscDbc/`.

### 🧹 Code Quality

- `OdbcError` chain: `std::vector<std::unique_ptr<OdbcError>>` replaces raw linked list.
- `IscConnection::statements`, `IscStatement::resultSets`, `IscBlob::blobs`: `std::vector` replaces `LinkedList`.
- `OdbcObject`, `OdbcError`, `OdbcEnv`: diagnostic fields made `private`/`protected`.
- `EnvShare`: Meyer's Singleton (construct-on-first-use) fixes static initialization order.
- Concrete `IscDbc` classes marked `final` for devirtualization.
- ~3,750 lines of dead code removed (Phase 13).
- ATL/DTC distributed transaction support removed entirely.
- ODBC escape sequence processing removed (SQL sent as-is to Firebird).

### 🗑️ What Was Removed

| Component | Reason |
|-----------|--------|
| `Builds/` (14 configs) | Replaced by CMake |
| `FBClient.Headers/` (60 files) | Fetched via CMake FetchContent |
| `Headers/` (SQL.H, SQLEXT.H) | System SDK headers used |
| `OdbcJdbcSetup/` (30 files) | GUI not maintained; use odbcconf/odbcinst |
| `Install/` (HTML help, InnoSetup) | Replaced by WiX MSI |
| `Res/` (5 locale files) | i18n not supported |
| `Test/`, `Tests/`, `JdbcTest/`, `TestInstall/` | Replaced by Google Test |
| `ChangeLog` files | Git history replaces these |
| ATL/DTC support | Not maintained; Firebird doesn't support XA/DTC |
| ODBC escape sequences | Legacy feature; SQL passthrough is standard practice |

---

## Testing

- **401 tests** across 34 Google Test suites.
- Tests run against a **real Firebird 5.0 database** (both locally and in CI).
- Test categories include: null handle safety, connection options, catalog functions, data type conversions (results and parameters), array binding, descriptors, cursors (forward-only and scrollable), error handling, ODBC 3.8 compliance, SQL_GUID, BLOBs, savepoints, multi-statement handles, query timeout, connection pool reset, Unicode (WCHAR), and more.
- Tests gracefully skip (`GTEST_SKIP()`) when no database connection is available.
- CI runs on Windows x64 and Linux x64/ARM64.

---

## Breaking Changes

- **Build system**: CMake is now required. Legacy build configurations (MSVC .sln, makefiles, MinGW, Borland) are removed.
- **Vendored headers**: Firebird headers are no longer in the repository. CMake fetches them at build time.
- **Source layout**: Source files moved from root to `src/` and `src/IscDbc/`.
- **CHARSET default**: When not specified in the connection string, `CHARSET` now defaults to `UTF8` (was `NONE`). This is a correctness improvement -- `CHARSET=NONE` with locale-dependent `mbstowcs` was producing incorrect results on non-UTF-8 locales.
- **ODBC escape sequences**: `{fn}`, `{d}`, `{ts}`, `{oj}` escape processing has been removed. SQL is sent as-is to Firebird. Applications using ODBC escapes should switch to native Firebird SQL syntax.
- **ATL/DTC**: Distributed transaction support (ATL COM) has been removed.
- **OdbcJdbcSetup GUI**: The setup dialog has been removed. Use `odbcconf` or `odbcinst` for driver registration.

---

## Transparency Note

This work was substantially assisted by AI (LLMs). I believe in being upfront about this: the AI was used for code generation, analysis, testing, and documentation throughout this project. Every change was reviewed, tested against a real Firebird database, and verified on CI. The detailed issue tracking in [Docs/FIREBIRD_ODBC_MASTER_PLAN.md](https://github.com/fdcastel/firebird-odbc-driver/blob/v4-pr/Docs/FIREBIRD_ODBC_MASTER_PLAN.md) documents the reasoning behind each change.

I believe the results speak for themselves: 401 passing tests, crash bugs eliminated, correct SQLSTATE mapping, functional Unicode on Linux, and measurable performance improvements. But I understand if the use of AI tooling is a concern, and I'm happy to discuss any specific changes in detail.


mrotteveel:
> OdbcJdbcSetup GUI: The setup dialog has been removed. Use odbcconf or odbcinst for driver registration.

Isn't this the thing that is shown in the ODBC Data Source Administrator in Windows? If so, it should remain in place.

fdcastel:
Thanks for the feedback, guys! I will address them soon. 👍🏻 

> Do you really need any additional test cases assuming you aleady have >400 tests?))))

Well... There's always room for improvement 😄

fdcastel:
> > Savepoint isolation: Per-statement SAVEPOINT/RELEASE/ROLLBACK TO SAVEPOINT when autoCommit=OFF. Failed statements don't corrupt the transaction.
> 
> This shouldn't be necessary. As far as I know, the engine already uses server-side savepoints to undo work on statement failure. 

Mark, you’re right. I’m also finding this statement very strange. I’ll investigate it. It seems like it may be a misinterpretation of what was done when savepoint support was added.

irodushka:
Please please

Do not touch the code that doesn't need to be touched) the code that works
and was fine tested about 2 years ago - see the link to the ticket above. I
mean escapes.

The best thing you can make here is creating a few additional tests around
it - that's welcome. But rewriting the code just for the sake of rewriting
it - not.

Regards



чт, 12 февр. 2026 г., 16:54 F.D.Castel ***@***.***>:

fdcastel:
> the code that works and was fine tested about 2 years ago - see the link to the ticket above. 

Sorry, what link / ticket?

Also, I’ll be in dire need of those tests, now. Reintegrating the old code at this point would probably be a no-go.

fdcastel:
I've been thinking about how we should move forward with this PR, and I'd like to propose the following approach.

As always, opinions, criticism, and suggestions are very welcome.

## General Strategy

* **Do not merge this PR.**
  Instead, we keep it open as a **reference for the intended end state** -- a kind of “target vision” for where we want to go.

* I will continue maintaining this branch (let's call it **“unofficial-v4-preview”**) for internal use and possibly for selected production scenarios.

  * This way, I'll be acting as the primary beta tester for it.
  * Of course, I'll provide unofficial builds for anyone interested in testing it.

* In parallel, and with @irodushka's guidance, I will start submitting **separate, well-scoped PRs**, each focused on a clearly defined feature or improvement (to be **agreed upon together**).

  * The goal is to incrementally achieve the same end result as this large PR -- **excluding the problematic decisions**(*) -- in a way that is easier to review, validate, and merge.

> (*) Some decisions in the current PR were mistakes (e.g., removing ODBC escape sequences and removing the ODBC Administrator UI forms). These should not be carried forward.

## Initial Breakdown

The following outlines an initial plan for how we should approach this problem. 

Each item will be submitted as a separate PR, potentially structured as a sequence of **dependent** commits (building on one another).

### 1. Build System

* Port the build system to CMake.
* Remove legacy build configurations (MSVC `.sln`, makefiles, MinGW, Borland).

### 2. Test Suite

* Port the test suite only. Many tests will initially fail.
  * That's acceptable. It will at least give us a clear target to work toward.
* Add any additional tests that @irodushka may have.
* Add ODBCQueryTool tests for escape sequences (and any others considered useful).

### 3. GitHub Actions automation

* Adopt GitHub Actions for build and tests
  - Ideally, also for releases -- though that decision belongs to @irodushka. It's not _required_ for now.

**The changes above are the most necessary.** Without them, the remaining tasks would require significant manual testing and verification effort, making progress inefficient and resource-intensive.


### 4. Next Steps: Following the Original Plan -- Without the Bad Parts (*)

From this point on, we largely follow the original roadmap, excluding the problematic decisions:

* **Phase 0:** [Stabilization (Crashes & Data Corruption)](https://github.com/fdcastel/firebird-odbc-driver/blob/fix-master-plan/Docs/FIREBIRD_ODBC_MASTER_PLAN.md#phase-0-stabilize-fix-crashes-and-data-corruption--completed--february-7-2026)

* **Phase 1:** [ODBC Spec Compliance (Error Mapping & Diagnostics)](https://github.com/fdcastel/firebird-odbc-driver/blob/v4-pr/Docs/FIREBIRD_ODBC_MASTER_PLAN.md#phase-1-odbc-spec-compliance-error-mapping--diagnostics)

* **Phase 2:** [Entry Point Hardening](https://github.com/fdcastel/firebird-odbc-driver/blob/fix-master-plan/Docs/FIREBIRD_ODBC_MASTER_PLAN.md#phase-2-entry-point-hardening--completed--february-7-2026)

* **Phase 4:** [Feature Completeness](https://github.com/fdcastel/firebird-odbc-driver/blob/fix-master-plan/Docs/FIREBIRD_ODBC_MASTER_PLAN.md#phase-4-feature-completeness--completed--february-7-2026)

* **Phase 7:** [ODBC Crusher-Identified Bugs](https://github.com/fdcastel/firebird-odbc-driver/blob/fix-master-plan/Docs/FIREBIRD_ODBC_MASTER_PLAN.md#phase-7-odbc-crusher-identified-bugs--completed--february-8-2026)

* **Phase 9:** [Firebird OO API Alignment & Simplification](https://github.com/fdcastel/firebird-odbc-driver/blob/fix-master-plan/Docs/FIREBIRD_ODBC_MASTER_PLAN.md#phase-9-firebird-oo-api-alignment--simplification)
  * At this stage, we may consider adopting `fb-cpp`.

* Adopt [vendored headers](https://github.com/fdcastel/firebird-odbc-driver/blob/fix-master-plan/Docs/FIREBIRD_ODBC_MASTER_PLAN.md#build-infrastructure-vendored-header-removal--completed--february-8-2026): Remove Firebird headers from the repository.
  - Configure CMake to fetch them at build time _(today)_; or
  - use `vcpkg`? _(better?)_

* Remove ATL / Rewrite distributed transaction support without ATL
  - Allow usage of distributed transactions in Linux/other non-Windows platforms, too.

* Port remaining Unicode-related changes.

* Port remaining performance improvements.

### 5. Final Cleanup (Very Last Step)
  * Reorganize source layout: Move source files from the repository root into `src/`.
  * Remove dead code.

irodushka:
Hi

@fdcastel - agree with you plan. This is exactly what I meant yesterday between the lines of my message%)
Except for one point - I think it's better to implement a test suite first. This way we get the basic point of testing coverage and success/failure stats AS IS, before any futher changes/fixes/enhancements are made. And, very important for me - we get a tool to check regress in the next steps.

What do you think about the following plan? -

### DevOps phase

So, in the **first step** we should get a test suite (with some test cases that are expected to fail) and it should be run manually.

The **2nd step** - a new build system. Well, I dont think removing the last actual sln (MsVc2022.win folder) is reasonable. All other *.win solutions - ok, they are outdated, but MsVc2022.win should be kept for the customers who prefer to build from the native Visual Studio sln.
+++ _Remove Firebird headers from the repository task_ - seems reasonable to include this in this step.
Testing: no regress and no enhancements are expected at this stage.

And the **3rd step** - a new CI (GitHib actions). At this stage, we move from manual launch of test suite to automated.
Testing: still, no regress and no enhancements are expected at this stage.

### Bugfix phase

Here we should fix the known errors and misbehavior cases. Every case must be issued as a separate ticket, branch and PR.
Testing: No regression, but improvements are expected - some of the previously failed tests should pass. If there's no test coverage for a given case, additional tests should be created.

### Improvement phase (new features, refactoring & optimizations)
[to do closer to the point]

fdcastel:
@irodushka First step done in #276.

Please review. I'll need it for the next steps.

fdcastel:
Superseded by #283. I’m closing this now. Work will continue in #282.

Issue #287, opened 2026-04-06
[EPIC] vNext Migration Plan


fdcastel:
**Source branch**: `v5-pr` ([PR #283](https://github.com/FirebirdSQL/firebird-odbc-driver/pull/283) — RFC only, DO NOT MERGE)  
**Target branch**: `master` (official)  
**Reference**: [`Docs/FIREBIRD_ODBC_MASTER_PLAN.original.md`](../tmp/v5-pr/Docs/FIREBIRD_ODBC_MASTER_PLAN.original.md) (v5-pr branch)  
**Last updated**: April 18, 2026

Legend:
| Status | Meaning |
|--------|---------|
| ✅ RESOLVED | Fix implemented and tested in `master` |
| 🔧 IN PROGRESS | Partially implemented or PR under review |
| ❌ OPEN | Not yet addressed |
| ⏯️ DEFERRED | delayed or put on hold until another task is finished |

---

## Already Merged to Master

These PRs have been completed and form the foundation for all subsequent work.

| # | Description | PR | Status |
|---|-------------|-----|--------|
| F-1 | **Google Test suite** — 375 tests (230 pass, 145 skip), regression testing baseline | [#276](https://github.com/FirebirdSQL/firebird-odbc-driver/pull/276) | ✅ RESOLVED |
| F-2 | **CMake build system** — Replaced 14 obsolete build configs, vendored headers fetched at build time, cross-platform (Windows + Linux + ARM64) | [#281](https://github.com/FirebirdSQL/firebird-odbc-driver/pull/281) | ✅ RESOLVED |
| F-3 | **GitHub Actions CI** — Build-and-test workflow (Windows + Linux), Firebird 5.0.2 via PSFirebird, 3 charset configurations, Invoke-Build script | [#282](https://github.com/FirebirdSQL/firebird-odbc-driver/pull/282) | ✅ RESOLVED |
| F-4 | **Array binding fix** — `sizeof(SQLINTEGER)` → `sizeof(SQLLEN)` indicator stride in OdbcStatement.cpp (fixes segfault on Linux) | [#279](https://github.com/FirebirdSQL/firebird-odbc-driver/pull/279) | ✅ RESOLVED |
[...]

PR #289, opened 2026-04-10
Add ASAN and Valgrind integration for CI test suite

### LINKED ISSUE #288: Adopt ASAN / Valgrind in the test suite
@fdcastel 

I think we should run all our tests 
1) with ASAN. That means we must make a special build (Debug, maybe wo extra logging) with -fsanitize-address, probably for both driver so & the test binary.
2) with Valgrind.

It will help us to maintain the code purity.

I just found a buffer overflow bug in legacy code (maybe it's already fixed in the futher PRs) while running tests with ASAN.

---

The best way is to add the custom cmake types (like DebugWithAsan & DebugWithValgrind), both with -O0, -g3 and no extra logging. It's ok to use simple Debug type for Valgrind run, if the extra logs aren't annoying. ASAN build is special due to the use of special flags & linkage.
And it's not a good idea to combine ASAN & Valgrind runs. So we should run our test suite twice. ASAN run has almost the same speed as the normal, Valgrind is much slower.

Perhaps it would be more efficient if I modify the CMake files and you handle the Windows installer?.. and of cause all that sanitizer stuff may be merged after this PR, no deps here. Do you expect any merge issues if I make changes to the master branch?

---

_Originally posted by @irodushka in https://github.com/FirebirdSQL/firebird-odbc-driver/issues/286#issuecomment-4224762569_

fdcastel:
## Summary

Integrate **AddressSanitizer (ASAN)** and **Valgrind** into the CI test pipeline so that memory bugs (buffer overflows, use-after-free, leaks, etc.) are caught automatically on every PR.

Closes #288

---

### CMake (`CMakeLists.txt`)

- **`ENABLE_ASAN`** option (default OFF, Linux/GCC/Clang only): appends `-fsanitize=address -fno-omit-frame-pointer` to compile and link flags. Suppresses `-DLOGGING` in Debug builds for cleaner sanitizer output.
- **`ENABLE_VALGRIND`** option (default OFF): finds the `valgrind` binary and configures `MEMORYCHECK_COMMAND` / `MEMORYCHECK_COMMAND_OPTIONS` for `ctest -T memcheck`.
- Mutual exclusion enforced: enabling both emits a `FATAL_ERROR`.
- Both options are guarded behind `if(NOT MSVC)` since MSVC ASAN has different semantics (future work).

### CMake Presets (`CMakePresets.json`)

- **`asan`** configure/build/test presets inheriting from `debug`, with `ASAN_OPTIONS` environment variable.
- **`valgrind`** configure/build/test presets inheriting from `debug`, with a 600s test timeout (Valgrind is ~10-20x slower).

### Build script (`firebird-odbc-driver.build.ps1`)

- New **`-Sanitizer`** parameter accepting `None`, `Asan`, `Valgrind`.
- Passes the corresponding `-DENABLE_ASAN=ON` or `-DENABLE_VALGRIND=ON` to CMake.
- Sets `ASAN_OPTIONS` / `LSAN_OPTIONS` environment variables at runtime for ASAN builds.
- Emits a warning and falls back to normal build on Windows.

### CI (`build-and-test.yml`)

- New **Linux x64 ASAN** matrix entry (ubuntu-22.04, Debug config).
- New **Linux x64 Valgrind** matrix entry (ubuntu-22.04, Debug config) with `valgrind` package installation.
- Sanitizer jobs set `ASAN_OPTIONS` and `LSAN_OPTIONS` environment variables.
- Sanitizer jobs do **not** upload release artifacts.

### Valgrind suppressions (`valgrind.supp`)

- Initial suppressions file seeded with a rule for `libfbclient.so` global initialization leaks.
- To be populated as false positives are discovered.

---

## Local usage

```bash
# ASAN build
cmake --preset asan
cmake --build --preset asan
ctest --preset asan

# Valgrind build
cmake --preset valgrind
cmake --build --preset valgrind
ctest --preset valgrind
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,14 +130,16 @@ else()
 
     # Compiler optimization flags (from PR #248 / old makefile.linux)
     add_compile_options(
-        "$<$<CONFIG:Debug>:-O0;-g3;-D_DEBUG;-DDEBUG;-fexceptions>"
+        "$<$<CONFIG:Debug>:-O0;-g3;-D_DEBUG;-fexceptions>"
         "$<$<CONFIG:Release>:-O3;-DNDEBUG;-ftree-loop-vectorize>"
         "$<$<CONFIG:RelWithDebInfo>:-O2;-g;-DNDEBUG>"
         "$<$<CONFIG:MinSizeRel>:-Os;-DNDEBUG>"
     )
 
-    # LOGGING: Debug builds only; sanitizer builds strip it for cleaner output.
+    # Debug macros: DEBUG and LOGGING are for Debug builds only;
+    # sanitizer builds strip them for cleaner output (less noise in reports).
     if(CMAKE_BUILD_TYPE STREQUAL "Debug" AND NOT BUILD_WITH_ASAN AND NOT BUILD_WITH_VALGRIND)
+        add_definitions(-DDEBUG)
         add_definitions(-DLOGGING)
     endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -137,11 +137,8 @@ else()
     )
 
     # LOGGING: Debug builds only; sanitizer builds strip it for cleaner output.
-    if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+    if(CMAKE_BUILD_TYPE STREQUAL "Debug" AND NOT BUILD_WITH_ASAN AND NOT BUILD_WITH_VALGRIND)
         add_definitions(-DLOGGING)
-    endif()
-    if(BUILD_WITH_ASAN OR BUILD_WITH_VALGRIND)
-        remove_definitions(-DLOGGING)
     endif()
 
     # SSE4.1 for x86/x86_64 (from PR #248)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,38 @@ option(BUILD_SHARED_LIBS "Build shared libraries" ON)
 option(BUILD_TESTING "Build tests" ON)
 
 # ---------------------------------------------------------------------------
+# Sanitizer options (Linux / GCC / Clang only)
+# ---------------------------------------------------------------------------
+if(NOT MSVC)
+    option(ENABLE_ASAN "Enable AddressSanitizer (-fsanitize=address)" OFF)
+    option(ENABLE_VALGRIND "Enable Valgrind memcheck via CTest" OFF)
+
+    if(ENABLE_ASAN AND ENABLE_VALGRIND)
+        message(FATAL_ERROR "ENABLE_ASAN and ENABLE_VALGRIND are mutually exclusive. "
+                            "ASAN instruments the binary at compile time; Valgrind instruments at runtime. "
+                            "They must not be combined.")
+    endif()
+
+    if(ENABLE_ASAN)
+        message(STATUS "AddressSanitizer: ENABLED")
+        add_compile_options(-fsanitize=address -fno-omit-frame-pointer)
+        add_link_options(-fsanitize=address)
+    endif()
+
+    if(ENABLE_VALGRIND)
+        find_program(VALGRIND_COMMAND valgrind)
+        if(NOT VALGRIND_COMMAND)
+            message(FATAL_ERROR "Valgrind not found but ENABLE_VALGRIND is ON. "
+                                "Install it with: sudo apt-get install valgrind")
+        endif()
+        message(STATUS "Valgrind memcheck: ENABLED (${VALGRIND_COMMAND})")
+        set(MEMORYCHECK_COMMAND ${VALGRIND_COMMAND})
+        set(MEMORYCHECK_COMMAND_OPTIONS
+            "--leak-check=full --error-exitcode=1 --suppressions=${CMAKE_SOURCE_DIR}/valgrind.supp")
+    endif()
+endif()
+
+# ---------------------------------------------------------------------------
 # CPU architecture detection (from PR #248)
 # ---------------------------------------------------------------------------
 # On MSVC multi-config generators, -A Win32/ARM64 sets CMAKE_GENERATOR_PLATFORM
@@ -94,8 +126,17 @@ else()
     set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
     # Compiler optimization flags (from PR #248 / old makefile.linux)
+    if(ENABLE_ASAN)
+        # ASAN builds: debug flags without extra logging noise
+        add_compile_options(
+            "$<$<CONFIG:Debug>:-O0;-g3;-D_DEBUG;-DDEBUG;-fexceptions>"
+        )
+    else()
+        add_compile_options(
+            "$<$<CONFIG:Debug>:-O0;-g3;-D_DEBUG;-DDEBUG;-DLOGGING;-fexceptions>"
+        )
+    endif()
     add_compile_options(
-        "$<$<CONFIG:Debug>:-O0;-g3;-D_DEBUG;-DDEBUG;-DLOGGING;-fexceptions>"
         "$<$<CONFIG:Release>:-O3;-DNDEBUG;-ftree-loop-vectorize>"
         "$<$<CONFIG:RelWithDebInfo>:-O2;-g;-DNDEBUG>"
         "$<$<CONFIG:MinSizeRel>:-Os;-DNDEBUG>"
@@ -249,8 +290,12 @@ endif()
 # Debug-specific definitions (matching .vcxproj: DEBUG;LOGGING for Debug configs)
 target_compile_definitions(OdbcFb PRIVATE
     $<$<CONFIG:Debug>:DEBUG>
-    $<$<CONFIG:Debug>:LOGGING>
 )
+if(NOT ENABLE_ASAN)
+    target_compile_definitions(OdbcFb PRIVATE
+        $<$<CONFIG:Debug>:LOGGING>
+    )
+endif()
 
 # ---------------------------------------------------------------------------
 # Testing
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,16 +131,18 @@ else()
     # Compiler optimization flags (from PR #248 / old makefile.linux)
     add_compile_options(
         "$<$<CONFIG:Debug>:-O0;-g3;-D_DEBUG;-DDEBUG;-fexceptions>"
-    )
-    # LOGGING definition: enable for Debug builds, suppress for sanitizer builds
-    if(NOT (BUILD_WITH_ASAN OR BUILD_WITH_VALGRIND))
-        add_definitions(-DLOGGING)
-    endif()
-    add_compile_options(
         "$<$<CONFIG:Release>:-O3;-DNDEBUG;-ftree-loop-vectorize>"
         "$<$<CONFIG:RelWithDebInfo>:-O2;-g;-DNDEBUG>"
         "$<$<CONFIG:MinSizeRel>:-Os;-DNDEBUG>"
     )
+
+    # LOGGING: Debug builds only; sanitizer builds strip it for cleaner output.
+    if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+        add_definitions(-DLOGGING)
+    endif()
+    if(BUILD_WITH_ASAN OR BUILD_WITH_VALGRIND)
+        remove_definitions(-DLOGGING)
+    endif()
 
     # SSE4.1 for x86/x86_64 (from PR #248)
     if(FBODBC_ARCH STREQUAL "x86" OR FBODBC_ARCH STREQUAL "i686"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,28 +42,31 @@ option(BUILD_SHARED_LIBS "Build shared libraries" ON)
 option(BUILD_TESTING "Build tests" ON)
 
 # ---------------------------------------------------------------------------
-# Sanitizer options (Linux / GCC / Clang only)
+# Sanitizer options
 # ---------------------------------------------------------------------------
-if(NOT MSVC)
-    option(ENABLE_ASAN "Enable AddressSanitizer (-fsanitize=address)" OFF)
-    option(ENABLE_VALGRIND "Enable Valgrind memcheck via CTest" OFF)
+option(BUILD_WITH_ASAN "Enable AddressSanitizer (-fsanitize=address)" OFF)
+option(BUILD_WITH_VALGRIND "Enable Valgrind memcheck via CTest" OFF)
 
-    if(ENABLE_ASAN AND ENABLE_VALGRIND)
-        message(FATAL_ERROR "ENABLE_ASAN and ENABLE_VALGRIND are mutually exclusive. "
-                            "ASAN instruments the binary at compile time; Valgrind instruments at runtime. "
-                            "They must not be combined.")
-    endif()
+if(BUILD_WITH_ASAN AND BUILD_WITH_VALGRIND)
+    message(FATAL_ERROR "BUILD_WITH_ASAN and BUILD_WITH_VALGRIND are mutually exclusive. "
+                        "ASAN instruments the binary at compile time; Valgrind instruments at runtime. "
+                        "They must not be combined.")
+endif()
 
-    if(ENABLE_ASAN)
+if(MSVC AND (BUILD_WITH_ASAN OR BUILD_WITH_VALGRIND))
+    message(WARNING "Sanitizers are not yet supported on MSVC. "
+                    "BUILD_WITH_ASAN / BUILD_WITH_VALGRIND will be ignored.")
+elseif(NOT MSVC)
+    if(BUILD_WITH_ASAN)
         message(STATUS "AddressSanitizer: ENABLED")
         add_compile_options(-fsanitize=address -fno-omit-frame-pointer)
         add_link_options(-fsanitize=address)
     endif()
 
-    if(ENABLE_VALGRIND)
+    if(BUILD_WITH_VALGRIND)
         find_program(VALGRIND_COMMAND valgrind)
         if(NOT VALGRIND_COMMAND)
-            message(FATAL_ERROR "Valgrind not found but ENABLE_VALGRIND is ON. "
+            message(FATAL_ERROR "Valgrind not found but BUILD_WITH_VALGRIND is ON. "
                                 "Install it with: sudo apt-get install valgrind")
         endif()
         message(STATUS "Valgrind memcheck: ENABLED (${VALGRIND_COMMAND})")
@@ -126,15 +129,12 @@ else()
     set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
     # Compiler optimization flags (from PR #248 / old makefile.linux)
-    if(ENABLE_ASAN)
-        # ASAN builds: debug flags without extra logging noise
-        add_compile_options(
-            "$<$<CONFIG:Debug>:-O0;-g3;-D_DEBUG;-DDEBUG;-fexceptions>"
-        )
-    else()
-        add_compile_options(
-            "$<$<CONFIG:Debug>:-O0;-g3;-D_DEBUG;-DDEBUG;-DLOGGING;-fexceptions>"
-        )
+    add_compile_options(
+        "$<$<CONFIG:Debug>:-O0;-g3;-D_DEBUG;-DDEBUG;-fexceptions>"
+    )
+    # LOGGING definition: enable for Debug builds, suppress for sanitizer builds
+    if(NOT (BUILD_WITH_ASAN OR BUILD_WITH_VALGRIND))
+        add_definitions(-DLOGGING)
     endif()
     add_compile_options(
         "$<$<CONFIG:Release>:-O3;-DNDEBUG;-ftree-loop-vectorize>"
@@ -284,16 +284,6 @@ else()
         odbcinst
         dl
         pthread
-    )
-endif()
-
-# Debug-specific definitions (matching .vcxproj: DEBUG;LOGGING for Debug configs)
-target_compile_definitions(OdbcFb PRIVATE
-    $<$<CONFIG:Debug>:DEBUG>
-)
-if(NOT ENABLE_ASAN)
-    target_compile_definitions(OdbcFb PRIVATE
-        $<$<CONFIG:Debug>:LOGGING>
     )
 endif()
 
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -36,7 +36,7 @@
             "displayName": "Debug + AddressSanitizer",
             "inherits": "debug",
             "cacheVariables": {
-                "ENABLE_ASAN": "ON"
+                "BUILD_WITH_ASAN": "ON"
             }
         },
         {
@@ -44,7 +44,7 @@
             "displayName": "Debug + Valgrind",
             "inherits": "debug",
             "cacheVariables": {
-                "ENABLE_VALGRIND": "ON"
+                "BUILD_WITH_VALGRIND": "ON"
             }
         }
     ],
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -30,6 +30,22 @@
             "cacheVariables": {
                 "CMAKE_BUILD_TYPE": "Debug"
             }
+        },
+        {
+            "name": "asan",
+            "displayName": "Debug + AddressSanitizer",
+            "inherits": "debug",
+            "cacheVariables": {
+                "ENABLE_ASAN": "ON"
+            }
+        },
+        {
+            "name": "valgrind",
+            "displayName": "Debug + Valgrind",
+            "inherits": "debug",
+            "cacheVariables": {
+                "ENABLE_VALGRIND": "ON"
+            }
         }
     ],
     "buildPresets": [
@@ -42,6 +58,16 @@
             "name": "debug",
             "configurePreset": "default",
             "configuration": "Debug"
+        },
+        {
+            "name": "asan",
+            "configurePreset": "asan",
+            "configuration": "Debug"
+        },
+        {
+            "name": "valgrind",
+            "configurePreset": "valgrind",
+            "configuration": "Debug"
         }
     ],
     "testPresets": [
@@ -50,6 +76,26 @@
             "configurePreset": "default",
             "output": {
                 "outputOnFailure": true
+            }
+        },
+        {
+            "name": "asan",
+            "configurePreset": "asan",
+            "output": {
+                "outputOnFailure": true
+            },
+            "environment": {
+                "ASAN_OPTIONS": "detect_leaks=1:halt_on_error=1:print_stats=1"
+            }
+        },
+        {
+            "name": "valgrind",
+            "configurePreset": "valgrind",
+            "output": {
+                "outputOnFailure": true
+            },
+            "execution": {
+                "timeout": 600
             }
         }
     ]
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -36,7 +36,8 @@
             "displayName": "Debug + AddressSanitizer",
             "inherits": "debug",
             "cacheVariables": {
-                "BUILD_WITH_ASAN": "ON"
+                "BUILD_WITH_ASAN": "ON",
+                "BUILD_WITH_VALGRIND": "OFF"
             }
         },
         {
@@ -44,7 +45,8 @@
             "displayName": "Debug + Valgrind",
             "inherits": "debug",
             "cacheVariables": {
-                "BUILD_WITH_VALGRIND": "ON"
+                "BUILD_WITH_VALGRIND": "ON",
+                "BUILD_WITH_ASAN": "OFF"
             }
         }
     ],
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -12,7 +12,9 @@
             "description": "Default preset — uses 'build/' as the binary directory so that all tools (command-line, Visual Studio, CLion, VS Code) share the same layout.",
             "binaryDir": "${sourceDir}/build",
             "cacheVariables": {
-                "CMAKE_INSTALL_PREFIX": "${sourceDir}/install"
+                "CMAKE_INSTALL_PREFIX": "${sourceDir}/install",
+                "BUILD_WITH_ASAN": "OFF",
+                "BUILD_WITH_VALGRIND": "OFF"
             }
         },
         {
@@ -20,7 +22,9 @@
             "displayName": "Release",
             "inherits": "default",
             "cacheVariables": {
-                "CMAKE_BUILD_TYPE": "Release"
+                "CMAKE_BUILD_TYPE": "Release",
+                "BUILD_WITH_ASAN": "OFF",
+                "BUILD_WITH_VALGRIND": "OFF"
             }
         },
         {
@@ -28,7 +32,9 @@
             "displayName": "Debug",
             "inherits": "default",
             "cacheVariables": {
-                "CMAKE_BUILD_TYPE": "Debug"
+                "CMAKE_BUILD_TYPE": "Debug",
+                "BUILD_WITH_ASAN": "OFF",
+                "BUILD_WITH_VALGRIND": "OFF"
             }
         },
         {
--- a/MainUnicode.cpp
+++ b/MainUnicode.cpp
@@ -38,6 +38,21 @@
 extern FILE	*logFile;
 using namespace OdbcJdbcLibrary;
 
+#ifndef _WINDOWS
+// SQLWCHAR-aware length (in SQLWCHAR units), safe on Linux where
+// sizeof(wchar_t) != sizeof(SQLWCHAR). Do NOT use wcslen() on SQLWCHAR
+// data on Linux — it reads two SQLWCHARs per wchar_t and runs off the end.
+static size_t sqlwcharLen( const SQLWCHAR *s )
+{
+	size_t n = 0;
+	if ( !s )
+		return 0;
+	while ( s[n] )
+		++n;
+	return n;
+}
+#endif
+
 #ifdef _WINDOWS
 extern UINT codePage; // from Main.cpp
 #endif
@@ -85,7 +100,7 @@ public:
 			if ( length == SQL_NTS )
 				lengthString = 0;
 			else if ( retCountOfBytes )
-				lengthString = length / sizeof(wchar_t);
+				lengthString = length / sizeof(SQLWCHAR);
 			else
 				lengthString = length;
 		}
@@ -135,13 +150,33 @@ public:
 					if ( len > 0 )
 						len--;
 #else
-					len = mbstowcs( (wchar_t*)unicodeString, (const char*)byteString, lengthString );
+					// SQLWCHAR is 2 bytes on Linux (unixODBC defines it as unsigned short),
+					// but wchar_t is 4 bytes, so mbstowcs((wchar_t*)unicodeString, ...)
+					// both corrupts the output and risks overflowing the caller's buffer.
+					// Widen byte-by-byte into SQLWCHAR units, matching what unixODBC's
+					// ansi_to_unicode_copy() does internally. This is correct for the
+					// ASCII-only error/state strings that reach this code path; non-ASCII
+					// input will be handled by the broader ConvertingString rewrite tracked
+					// in issue #287 (Tier 9.1).
+					{
+						const SQLCHAR *src = byteString;
+						size_t i = 0;
+						while ( i < (size_t)lengthString && src[i] != 0 )
+						{
+							unicodeString[i] = (SQLWCHAR)( src[i] & 0xFF );
+							++i;
+						}
+						len = i;
+					}
 #endif
 				}
 
 				if ( len > 0 )
 				{
-					*(LPWSTR)(unicodeString + len) = L'\0';
+					// NUL-terminate in SQLWCHAR units. LPWSTR assignment of L'\0' writes
+					// sizeof(wchar_t) bytes, which overruns the output buffer by 2 bytes
+					// on Linux.
+					unicodeString[len] = 0;
 
 					if ( realLength )
 					{
@@ -170,12 +205,18 @@ public:
 		wchar_t saveWC;
 
 		if ( length == SQL_NTS )
+#ifdef _WINDOWS
 			length = (int)wcslen( (const wchar_t*)wcString );
-		else if ( wcString[length] != L'\0' )
+#else
+			length = (int)sqlwcharLen( wcString );
+#endif
+		else if ( wcString[length] != 0 )
 		{
 			ptEndWC = (wchar_t*)&wcString[length];
 			saveWC = *ptEndWC;
-			*ptEndWC = L'\0';
+			// Write a SQLWCHAR-sized NUL so we don't overrun the input by 2 bytes
+			// on Linux (wchar_t is 4 bytes there).
+			wcString[length] = 0;
 		}
 
 		if ( connection )
@@ -185,7 +226,10 @@ public:
 #ifdef _WINDOWS
 			bytesNeeded = WideCharToMultiByte( codePage, (DWORD)0, wcString, length, NULL, (int)0, NULL, NULL );
 #else
-			bytesNeeded = wcstombs( NULL, (const wchar_t*)wcString, length );
+			// See the symmetric comment in the destructor above: wcstombs assumes
+			// wchar_t-sized input, which corrupts SQLWCHAR data on Linux. The
+			// byte-narrowing loop below produces exactly `length` output bytes.
+			bytesNeeded = (size_t)length;
 #endif
 		}
 
@@ -198,7 +242,15 @@ public:
 #ifdef _WINDOWS
 			bytesNeeded = WideCharToMultiByte( codePage, 0, wcString, length, (LPSTR)byteString, (int)bytesNeeded, NULL, NULL );
 #else
-			bytesNeeded = wcstombs( (char *)byteString, (const wchar_t*)wcString, bytesNeeded );
+			{
+				size_t i = 0;
+				while ( i < (size_t)length && wcString[i] != 0 )
+				{
+					byteString[i] = (SQLCHAR)( wcString[i] & 0xFF );
+					++i;
+				}
+				bytesNeeded = i;
+			}
 #endif
 		}
 
@@ -219,16 +271,8 @@ protected:
 		case BYTESCHARS:
 			if ( lengthString )
 			{
-				// Floor the internal buffer at 8 bytes so that callers which pass a
-				// small SQLWCHAR output buffer (e.g. SQLGetDiagRecW with a 12-byte
-				// SQL state, yielding lengthString=3 on Linux where sizeof(wchar_t)=4)
-				// still have room for the 6-byte SQL state ("HY000\0") that
-				// OdbcError::sqlGetDiagRec strcpy's into this buffer. Keeping
-				// lengthString itself unchanged preserves the mbstowcs writeback
-				// bound and avoids smashing the caller's stack buffer.
-				const size_t bufSize = (lengthString + 2 < 8) ? 8 : (size_t)lengthString + 2;
-				byteString = new SQLCHAR[ bufSize ];
-				memset(byteString, 0, bufSize);
+				byteString = new SQLCHAR[ lengthString + 2 ];
+				memset( byteString, 0, lengthString + 2 );
 			}
 			else
 				byteString = NULL;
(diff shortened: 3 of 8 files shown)


fdcastel:
Yay! They are working!!!

image

fdcastel:
## Bug found and fixed by ASAN

The ASAN CI job introduced by this PR immediately caught a real memory bug in the driver.

### What ASAN reported

The first ASAN run failed on `DataTypeTest.SmallintRoundTrip` with:

```
AddressSanitizer: heap-buffer-overflow
WRITE of size 6 at ... (0 bytes past end of 5-byte region)
  #0 __interceptor_strcpy
  #1 OdbcError::sqlGetDiagRec
  #2 SQLGetDiagRecW
```

The full output is in the CI job log: GitHub Actions -> **Build and Test** -> job `ubuntu-22.04 / Debug / Asan` -> step **Run build script**.

### Root cause

`SQLGetDiagRecW` allocates a narrow-char work buffer for the SQL state:

```cpp
ConvertingString<> State( 12, sqlState );
```

In `ConvertingString`'s constructor, the byte-count to char-count conversion was:

```cpp
lengthString = length / sizeof(wchar_t);   // BUG
```

On Windows `sizeof(wchar_t) == sizeof(SQLWCHAR) == 2`, so `12/2 = 6` -> 8-byte buffer, no overflow.  
On Linux `sizeof(wchar_t) == 4` but `sizeof(SQLWCHAR) == 2` (unixODBC defines `SQLWCHAR` as `unsigned short`), so `12/4 = 3` -> 5-byte buffer.

`OdbcError::sqlGetDiagRec` then copies the SQL state ("HY000\0", 6 bytes) into that 5-byte buffer via `strcpy` -> 1-byte heap overflow. The same latent bug exists in `SQLErrorW` (same `State(12, sqlState)` pattern).

The bug was silently harmless on Windows and went undetected for years.

### Fix

One-line change in `MainUnicode.cpp` (commit 636972b):

```cpp
// Before
lengthString = length / sizeof(wchar_t);
// After
lengthString = length / sizeof(SQLWCHAR);
```

`sizeof(SQLWCHAR) == 2` on all platforms, giving a correct 8-byte buffer on Linux.

### CI after fix

All 4 matrix jobs now pass with no `continue-on-error` guard:

| Job | Result |
|-----|--------|
| ubuntu-22.04 / Release | pass |
| ubuntu-22.04 / Debug / ASAN | pass (375/375 tests) |
| ubuntu-22.04 / Debug / Valgrind | pass |
| windows-latest / Release | pass |

This is a concrete demonstration that ASAN integration pays off immediately.

fdcastel:
@irodushka Enjoy! 😉

irodushka:
Hi @fdcastel 

Can you please resolve the conflicts after merging PR#286
I'll look at this PR after Orthodox Easter)

Cheers!

fdcastel:
> Can you please resolve the conflicts after merging PR#286 I'll look at this PR after Orthodox Easter)

Sure! I’ll look into this in a few hours.

fdcastel:
@irodushka Rebased onto current `master`.

Conflicts were in `build-and-test.yml` and `firebird-odbc-driver.build.ps1`, caused by the expanded matrix (Windows ARM64, Windows x86/WoW64, Linux ARM64, multi-Firebird-version runs) introduced by master. 

Resolution: kept all new upstream matrix entries and added the sanitizer entries on top; merged the `Architecture` and `Sanitizer` params in the build script so both features coexist.

CMakeLists.txt
```diff
@@ -41,6 +41,38 @@ set(CMAKE_C_STANDARD_REQUIRED ON)
 option(BUILD_SHARED_LIBS "Build shared libraries" ON)
 option(BUILD_TESTING "Build tests" ON)
 
+# ---------------------------------------------------------------------------
+# Sanitizer options (Linux / GCC / Clang only)
+# ---------------------------------------------------------------------------
+if(NOT MSVC)
+    option(ENABLE_ASAN "Enable AddressSanitizer (-fsanitize=address)" OFF)
```
irodushka: I see no reason to put the options declaration in the `if(NOT MSVC)`. And we can emit a warning if running on Windows and the option is set - "temporarily not supported" or smth.

CMakeLists.txt
```diff
@@ -41,6 +41,38 @@ set(CMAKE_C_STANDARD_REQUIRED ON)
...
+    option(ENABLE_VALGRIND "Enable Valgrind memcheck via CTest" OFF)
```
irodushka: About the naming - we already have options with the BUILD prefix, so let's follow this style. 

BUILD_WITH_ASAN, BUILD_WITH_VALGRIND.

CMakeLists.txt
```diff
@@ -249,8 +290,12 @@ endif()
 # Debug-specific definitions (matching .vcxproj: DEBUG;LOGGING for Debug configs)
 target_compile_definitions(OdbcFb PRIVATE
     $<$<CONFIG:Debug>:DEBUG>
-    $<$<CONFIG:Debug>:LOGGING>
```
irodushka, line 252: Hmmm, don't understand the purpose of this code. Isn't it a duplicate logic, previously defined above?..

"$<$<CONFIG:Debug>:-O0;-g3;-D_DEBUG;-**DDEBUG**;-**DLOGGING**;-fexceptions>"

CMakeLists.txt
```diff
@@ -94,8 +126,17 @@ else()
     set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
     # Compiler optimization flags (from PR #248 / old makefile.linux)
+    if(ENABLE_ASAN)
```
irodushka: Don't like this logic.

-DLOGGING is even not an option, it is a definition, so we can easily remove it with the `remove_definitions()` cmake function. So we make a solid build-type-options matrix and then remove the LOGGING definition for ASAN build.

Another question - why only for ASAN?.. Valgrind build too, I suppose?..

CMakeLists.txt
fdcastel: Good call — addressed in 84b9dec. The `option()` declarations are now at the top level and a `message(WARNING ...)` is emitted when the options are set on MSVC.

CMakeLists.txt
fdcastel: Agreed — renamed to `BUILD_WITH_ASAN` / `BUILD_WITH_VALGRIND` in 84b9dec.

CMakeLists.txt
fdcastel: You're right — that was duplicate logic left over from an earlier iteration. The duplicate `target_compile_definitions(OdbcFb ...)` block was removed in 84b9dec; `-DDEBUG` (and previously `-DLOGGING`) were already being added via `add_compile_options` in the Linux section.

CMakeLists.txt
fdcastel: Good points. Restructured in e28f255:

- The Linux build-type options matrix is now a single consolidated `add_compile_options(...)` block.
- `LOGGING` is expressed via `add_definitions(-DLOGGING)` for Debug, then `remove_definitions(-DLOGGING)` is applied for sanitizer builds — the add/remove pattern you suggested.
- Applied to both `BUILD_WITH_ASAN` and `BUILD_WITH_VALGRIND` (earlier commit 84b9dec already extended the logic to Valgrind).

As a side benefit, this also fixed a bug introduced in the previous iteration where `LOGGING` was being added unconditionally (not scoped to Debug), which broke the Linux Release build with `LOG_FILE was not declared in this scope`.

fdcastel:
@irodushka Thanks for the valuable insights.

I’ve just pushed all the changes. Please take a look when you have a moment.

CMakeLists.txt
```diff
@@ -95,12 +130,20 @@ else()
...
     add_compile_options(
-        "$<$<CONFIG:Debug>:-O0;-g3;-D_DEBUG;-DDEBUG;-DLOGGING;-fexceptions>"
+        "$<$<CONFIG:Debug>:-O0;-g3;-D_DEBUG;-DDEBUG;-fexceptions>"
         "$<$<CONFIG:Release>:-O3;-DNDEBUG;-ftree-loop-vectorize>"
         "$<$<CONFIG:RelWithDebInfo>:-O2;-g;-DNDEBUG>"
         "$<$<CONFIG:MinSizeRel>:-Os;-DNDEBUG>"
     )
 
+    # LOGGING: Debug builds only; sanitizer builds strip it for cleaner output.
+    if(CMAKE_BUILD_TYPE STREQUAL "Debug")
```
irodushka: Mmmm... Why not leave -DLOGGING in the <CONFIG:Debug> section (no changing that line at all) and simply do
```
if(BUILD_WITH_ASAN OR BUILD_WITH_VALGRIND)
        remove_definitions(-DLOGGING)
endif()
````
?

Or you want to say that the definition LOGGING, added by add_compile_options() cmd, cannot be removed by remove_definitions()? In that case we can reduce this code to
```
if(CMAKE_BUILD_TYPE STREQUAL "Debug" AND NOT BUILD_WITH_ASAN AND NOT BUILD_WITH_VALGRIND)
    add_definitions(-DLOGGING)
endif()
```

Comment (fdcastel) [reply CMakeLists.txt]: Exactly right — `remove_definitions()` can only undo `add_definitions()`, not definitions baked into `add_compile_options()` generator expressions. Simplified to the single combined condition you suggested in 52f0495.

**Comment by irodushka** [comment]
@fdcastel 

We have a bug!
```
-- Git describe: 52f0495
CMake Warning at cmake/GetVersionFromGit.cmake:75 (message):
  Could not parse git describe output: 52f0495
Call Stack (most recent call first):
  CMakeLists.txt:15 (include)


CMake Warning at cmake/GetVersionFromGit.cmake:76 (message):
  Using fallback version 0.0.0.0
Call Stack (most recent call first):
  CMakeLists.txt:15 (include)
```

The reason is in the `--exclude "*-*"` option - 
`git describe --tags --match "v[0-9]*.[0-9]*.[0-9]*" --long --always`  returns `v3.5.0-rc1-8-g52f0495`, but `git describe --tags --match "v[0-9]*.[0-9]*.[0-9]*" --long --always --exclude "*-*"`  returns just the tail `52f0495`


Can you please remind me what do you expect specifying the `--exclude` option here?.. And what's the purpose of the logic - first try with the --exclude, and if we got an error (GIT_DESCRIBE_RESULT==0) try without --exclude?
Obviously this logic is not working.

**Comment by fdcastel** [comment]
@irodushka you're right, good catch. Fixed in b7aa955.

**Intent of the logic**
- `--match "v[0-9]*.[0-9]*.[0-9]*"` restricts candidates to semver tags, skipping legacy `v1-1-release` etc.
- `--exclude "*-*"` was meant to further restrict to **stable** `vX.Y.Z` tags (excluding prereleases like `v3.5.0-rc5`) so that a local build would be numbered off the latest stable release.
- If there is no stable `vX.Y.Z` tag yet (currently the case — every tag in the repo has a prerelease suffix), the second call drops `--exclude` and picks up the prerelease tag. The parser regex already handles the optional `-<prerelease>` suffix.

**Why it was broken**
- `--always` on the first call tells `git describe` to fall back to the short commit hash when nothing matches, and in that case the command exits **0**. So `GIT_DESCRIBE_RESULT` was `0`, the `if(NOT ... EQUAL 0)` branch never fired, and `GIT_DESCRIBE_OUTPUT` ended up as just `52f0495`, which then failed to match the parser regex → fallback to 0.0.0.0.

**The fix**
- Drop `--always` from both calls. The first call now exits non-zero when every tag is excluded, correctly triggering the fallback. If the fallback also finds no tags, the existing `message(WARNING "git describe failed")` path reports it cleanly.

Verified locally: `-- Git describe: v3.5.0-rc5-10-g52f0495` / `-- Firebird ODBC Driver version: 3.5.0.11`. All 11 CI jobs green on [run 24527813786](https://github.com/FirebirdSQL/firebird-odbc-driver/actions/runs/24527813786).

**Comment by irodushka** [comment]
Aha. Great)

Another note - let's exclude `-DDEBUG` from ASAN/Valgrind builds too, this macro adds some extra logging we actually don't need

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 748b82b..0769b1d 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,7 +130,7 @@ else()
 
     # Compiler optimization flags (from PR #248 / old makefile.linux)
     add_compile_options(
-        "$<$<CONFIG:Debug>:-O0;-g3;-D_DEBUG;-DDEBUG;-fexceptions>"
+        "$<$<CONFIG:Debug>:-O0;-g3;-D_DEBUG;-fexceptions>"
         "$<$<CONFIG:Release>:-O3;-DNDEBUG;-ftree-loop-vectorize>"
         "$<$<CONFIG:RelWithDebInfo>:-O2;-g;-DNDEBUG>"
         "$<$<CONFIG:MinSizeRel>:-Os;-DNDEBUG>"
@@ -139,6 +139,7 @@ else()
     # LOGGING: Debug builds only; sanitizer builds strip it for cleaner output.
     if(CMAKE_BUILD_TYPE STREQUAL "Debug" AND NOT BUILD_WITH_ASAN AND NOT BUILD_WITH_VALGRIND)
         add_definitions(-DLOGGING)
+        add_definitions(-DDEBUG)
     endif()
 
     # SSE4.1 for x86/x86_64 (from PR #248)
```

**Comment by irodushka** [comment]
@fdcastel 

Bad thing happens.
Your fix 

```
// Before
lengthString = length / sizeof(wchar_t);
// After
lengthString = length / sizeof(SQLWCHAR);
```

is not good. I get
```
...
SQLGetDiagRecW
SQLGetDiagRecW
*** stack smashing detected ***: terminated
Aborted (core dumped)
```

while running `./firebird_odbc_tests`, every time, with ASAN, without ASAN, doesn't matter.
Without the fix (`lengthString = length / sizeof(wchar_t);`) the test suite runs OK but obviously the ASAN build is failing.
I wonder why don't you see this and why we don't see this in the CI logs.

So it's not such a trivial case with that bug.

I will investigate tomorrow for it's a bit late here now. But you understand, I'm holding off on this PR until we solve this mystery.

**Comment by fdcastel** [comment]
Thanks @irodushka -- I will review all of them all in a few hours.

**Comment by fdcastel** [comment]
Done in 5323214 — moved both -DDEBUG and -DLOGGING to the sanitizer guard. All 11 CI jobs green.

**Comment by fdcastel** [comment]
The wchar_t → SQLWCHAR fix (commit 832d8e7) is correct: on Linux unixODBC defines SQLWCHAR=2 bytes but wchar_t=4 bytes, causing the buffer underallocation you caught with ASAN. That's a real bug fix, not a bug.

The stack smashing you're seeing with the tests is a separate issue — likely the test itself or something in the test harness. The heap-buffer-overflow in SQLGetDiagRecW that the wchar fix addresses should actually be *resolved* by it. Can you isolate which test triggers the stack smashing and whether it's reproducible without ASAN?

**Comment by irodushka** [comment]
> The wchar_t → SQLWCHAR fix (commit [832d8e7](https://github.com/FirebirdSQL/firebird-odbc-driver/commit/832d8e7a1592edc0d8665228e08cfcc34cc8edc3)) is correct: on Linux unixODBC defines SQLWCHAR=2 bytes but wchar_t=4 bytes, causing the buffer underallocation you caught with ASAN. That's a real bug fix, not a bug.

I don't argue with that. I only want to say, ~~if there is a way,~~ this undoubtly real bug fix seems to be incomplete)
At least - in MainUnicode.cpp:1166 we have just the same construction https://github.com/fdcastel/firebird-odbc-driver/blob/532321435a3ed4a3e34f2a0963a947e60084925d/MainUnicode.cpp#L1166. I tryed to change it to SQLWCHAR but unfortunately it has no effect.


> The stack smashing you're seeing with the tests is a separate issue — likely the test itself or something in the test harness. The heap-buffer-overflow in SQLGetDiagRecW that the wchar fix addresses should actually be _resolved_ by it. Can you isolate which test triggers the stack smashing and whether it's reproducible without ASAN?

The very first test case  - DataTypeTest.SmallintRoundTrip. It is easily reproducible wo ASAN as well.
The crash occures right during the DataTypeTest::SetUp() inside the std::make_unique<> ctor - 100% the stack is already broken at that moment.

```
(gdb) 
70	    void SetUp() override {
(gdb) 
98	    }
(gdb) 
DataTypeTest::SetUp (this=0x604000005350) at /home/irod/code/firebird-odbc-driver/firebird-odbc-driver/tests/test_data_types.cpp:11
11	        if (::testing::Test::IsSkipped()) return;
(gdb) 
13	        table_ = std::make_unique<TempTable>(this, "ODBC_TEST_TYPES",
(gdb) 
*** stack smashing detected ***: terminated

Thread 1 "firebird_odbc_t" received signal SIGABRT, Aborted.
```

I'll take a look when I've sorted out my affairs at my main job.

P.S. Actually not in the std::make_unique<> ctor. This ctor calls TempTable ctor, then ExecIgnoreError() -> unixodbc.SQLExecDirect() -> extract_diag_error_w() -> stack guard is fired.

**Comment by irodushka** [comment]
And, @fdcastel, I'll let you in on a secret!)

I have a customer in Moscow, he's using the ODBC driver in his project (on Linux), and he builds it with ASAN. He was the first one who got that ASAN error with buffer overflow, so when I asked you to implement the ASAN build I really knew all this story in advance) I fixed it in a fast QuickWin way for him (locally) - just changed the strcpy() calls to strncpy() in a few places and hardcoded the length limit. Ugly but efficient.

Of cause your fix is much more elegant, preferable and the right direction anyway, no doubt here.

**Comment by fdcastel** [comment]
> I fixed it in a fast QuickWin way for him (locally) - just changed the strcpy() calls to strncpy() in a few places and hardcoded the length limit. Ugly but efficient.

Hahahaha Fantastic!

- first make it work
- then make it right
- finally, make it fast.

File: MainUnicode.cpp
Comment (irodushka) [inline MainUnicode.cpp:138]: This place is terrible. The root of evil is here.
`mbstowcs` will convert each byte from byteString to FOUR bytes (sizeof(wchar_t)) in unicodeString. Given that unicodeString is represented in SQLWCHAR units, this inevitably leads to disaster. 


```suggestion
					//len = mbstowcs( (wchar_t*)unicodeString, (const char*)byteString, lengthString );

					mbstate_t state = {0};
                    const char *p = (const char*)byteString;
                    size_t out_idx = 0;
                    size_t rc;
                    char16_t *u16 = (char16_t *)unicodeString;

                    while ( (rc = std::mbrtoc16(&u16[out_idx], p, lengthString, &state)) )
                    {
                        if (rc == std::size_t(-1) || rc == std::size_t(-2))
                            break; // Invalid or truncated input
                        else if (rc == std::size_t(-3))
                            out_idx++; // UTF-16 high surrogate
                        else
                        {
                            p += rc;
                            out_idx++;
                        }
                    }
                    len = out_idx;
```

irodushka: I think this code will work just fine on Windows as well, so maybe we have no need to gate it with #ifdef here. I haven't tested it on Windows, though. On Linux, it seems to work fine, both with and wo ASAN and Valgrind.

fdcastel: Changes made:
- Implemented the proposed direction: Linux conversion now uses `std::mbrtoc16` and writes directly to SQLWCHAR-compatible storage.
- Added handling for `mbrtoc16` return cases (`-1`, `-2`, `-3`, `0`) and bounded output.
- Kept `#ifdef _WINDOWS` split intentionally.

I applied essentially this fix in the Linux branch: switched from `mbstowcs((wchar_t*)...)` to `std::mbrtoc16` with explicit output bounds and safe SQLWCHAR termination. This removes the wchar_t/SQWCHAR size mismatch at the exact reviewed site. 

P.S.: I kept the platform split intentionally. On Windows we currently must preserve `codePage`-driven conversion (`MultiByteToWideChar(codePage, ...)`). A unified `mbrtoc16` path would be locale-driven and could change behavior for non-UTF8 code pages. So I applied `mbrtoc16` in the Linux branch (where it fixes the SQLWCHAR mismatch) and kept Windows on the existing explicit code-page API for compatibility.

fdcastel:
Regarding [this](https://github.com/FirebirdSQL/firebird-odbc-driver/pull/289#issuecomment-4266851445) (*) comment:

Yep! The fix was incomplete in two places. 🤦🏻‍♂️ 

I pushed a follow-up that removes the remaining `wchar_t`-sized assumptions in `MainUnicode.cpp` for the investigated path:
- Linux bytes->SQLWCHAR conversion now uses `std::mbrtoc16` (no `mbstowcs((wchar_t*)...)` cast), with bounded output and safe SQLWCHAR null termination.
- `SQLSetDescFieldW` now computes character count using `sizeof(SQLWCHAR)`.
- I also replaced the remaining `wcslen((wchar_t*)...)` use in the touched flow with a SQLWCHAR-aware length helper.
This directly addresses the stack-corruption risk you described in the `SQLGetDiagRecW` path." 

Changes made:
- Replaced the Linux `mbstowcs` branch with UTF-16 aware conversion using `std::mbrtoc16` into `SQLWCHAR` storage.
- Added explicit output-capacity guard to reserve room for trailing `\0`.
- Changed null termination write from `*(LPWSTR)(unicodeString + len)` to `unicodeString[len] = 0`.
- Fixed `SQLSetDescFieldW` length split to use `sizeof(SQLWCHAR)`.
- Removed additional `wcslen((wchar_t*)...)` assumptions in touched paths by introducing `SqlWcharLength(const SQLWCHAR*)` and using it where relevant in `MainUnicode.cpp`.

_**(*) TIP:** Prefer GitHub Review mode instead of regular comments whenever possible: it's easier to keep each remark organized in its own thread._

irodushka:
Don't like it. It seems that  `std::mbrtoc16 ` is not a good choice for it may produce variable-length UTF16, and the ODBC needs fixed-length 2-byte encoding, not UTF-16 but UCS-2 in fact... 

From the other hand, if we interpret rc == -3 as an error, and AFAIK the case with the surrogate pair can happen only if the source string contains smth more than ANSI - I mean UTF8 with >1 bytes per character - maybe it will be ok.

And maybe it's better to use iconv here.

One thing I'm sure of - that `ConvertingString` class is awful) and must be totally rewritten.
 
@fdcastel do we have a dedicated test for the wide-char ODBC calls? The infamous `SQLGetDiagRecW` would by fine.




P.S. I'm looking into the unixODBC sources and they use iconv, or the naive approach if iconv is not installed:

```
SQLWCHAR *ansi_to_unicode_copy( SQLWCHAR * dest, char *src, SQLINTEGER buffer_len, DMHDBC connection, int *wlen )
{
...

    for ( i = 0; i < buffer_len && src[ i ] != 0; i ++ )
    {
#ifdef SQL_WCHART_CONVERT
        dest[ i ] = src[ i ] & 0x000000ff;
#else
        dest[ i ] = src[ i ] & 0x00FF;
#endif
    }

    if ( wlen ) 
    {
        *wlen = i;
    }

    dest[ i ] = 0;

    return dest;
}
```

fdcastel:
Fair points:

- **On `mbrtoc16` vs UCS-2/UTF-16:** You're right that the added complexity isn't justified for this path. 
  - The data in `byteString` at the diagnostic write-back point is always ASCII: SQL states ("HY000", "42S22" etc.) and Firebird error messages. I will replace `mbrtoc16` with the same simple byte-widening loop unixODBC uses `dest[i] = src[i] & 0x00FF`  which is correct, locale-independent, and considerably simpler.
  - For a general-purpose SQLWCHAR converter that handles non-ASCII input, I agree that `iconv` is the proper answer, but that belongs in a broader `ConvertingString` redesign (see below).

- **On the surrogate pair handling being "OK if treating `rc==-3` as error":** You CAN treat `-3` as error safely in practice because surrogates only arise for code points above U+FFFF, which never appear in ASCII error messages. 
  - But I will replace the whole block anyway.

- **On `iconv`:** Agreed it is the right long-term answer for `ConvertingString`. 
  - But adding it in this PR would be scope creep. I'd be happy to open an issue for it.

- **On `ConvertingString` being awful:** FULLY agreed. 
  - The destructor-as-converter pattern is confusing, error paths are silent, and the `wchar_t`/`SQLWCHAR` confusion is in both directions (I found the same `wcstombs((wchar_t*)wcString, ...)` bug in `convUnicodeToString` too).
  - This deserves a dedicated rewrite PR.

- **On a `SQLGetDiagRecW` test:** No such test exists. 
  - I'll add `GetDiagRecW` to `test_errors.cpp` in this commit. 
  - It will directly exercises the stack-smash path you found and will catch any regression.

I cannot work on the above point now. Feel free to add any comments or insights before I start to work on them.

MainUnicode.cpp
```diff
@@ -135,13 +148,42 @@ class ConvertingString
...
+
+					len = outCount;
 #endif
 				}
 
-				if ( len > 0 )
+				if ( lengthString > 0 )
 				{
-					*(LPWSTR)(unicodeString + len) = L'\0';
+					unicodeString[len] = (SQLWCHAR)0;
```
irodushka: What is the point of touching  this line?

`typedef WCHAR* 	            LPWSTR;`
`typedef  WCHAR         SQLWCHAR;`

So it's nothing wrong or bad here.

And mention - you've changed access style from `&out[outCount]` to `out + outCount` here
						size_t rc = std::mbrtoc16( **out + outCount**, p, inRemaining, &state );
but are still using access [] here.

Looks like a style mix-up.

MainUnicode.cpp
```diff
@@ -135,13 +148,42 @@ class ConvertingString
...
+						inRemaining -= rc;
+						++outCount;
+					}
+
+					len = outCount;
 #endif
 				}
 
-				if ( len > 0 )
+				if ( lengthString > 0 )
```
irodushka: What for?

MainUnicode.cpp
```diff
@@ -135,13 +148,42 @@ class ConvertingString
 					if ( len > 0 )
 						len--;
 #else
-					len = mbstowcs( (wchar_t*)unicodeString, (const char*)byteString, lengthString );
+					// SQLWCHAR is UTF-16 on Linux; convert multibyte input directly into SQLWCHAR units.
+					mbstate_t state = mbstate_t();
+					const char* p = (const char*)byteString;
+					size_t inRemaining = (size_t)lengthString;
```
irodushka: Incorrect. The bytes count must include the terminating 0 - see mbrtoc16 usage (https://en.cppreference.com/cpp/string/multibyte/mbrtoc16)

So it must be lengthString + 1, for byteString is allocated as `new SQLCHAR[ lengthString + 2 ];` (and do not ask me why +2)

Otherwise the `rc` will never be set to 0 cause the mbrtoc16 will never access the last zero byte.

MainUnicode.cpp
```diff
@@ -135,13 +148,42 @@ class ConvertingString
...
+					char16_t* out = (char16_t*)unicodeString;
+					size_t outCount = 0;
+					const size_t outCapacity = (lengthString > 0) ? (size_t)lengthString - 1 : 0;
+
+					while ( inRemaining > 0 && outCount < outCapacity )
```
irodushka: Please avoid such a comparison of unsigned vars. `inRemaining != 0` or just `inRemaining`. Unsigned can  never be <0. And such errors (if an overflow occurs) are very difficult to detect.

MainUnicode.cpp
```diff
@@ -135,13 +148,42 @@ class ConvertingString
...
 						len--;
 #else
-					len = mbstowcs( (wchar_t*)unicodeString, (const char*)byteString, lengthString );
+					// SQLWCHAR is UTF-16 on Linux; convert multibyte input directly into SQLWCHAR units.
+					mbstate_t state = mbstate_t();
+					const char* p = (const char*)byteString;
+					size_t inRemaining = (size_t)lengthString;
+					char16_t* out = (char16_t*)unicodeString;
+					size_t outCount = 0;
+					const size_t outCapacity = (lengthString > 0) ? (size_t)lengthString - 1 : 0;
```
irodushka: I don't see any point in this code and the check below. Looks like the redundant condition. The lengthString is _calculated_ on the base of the length, declared for the unicodeString, so what's the purpose is to check the out capacity if we're already checking the input remaining?

And why `(size_t)lengthString - 1`? What if the lengthString == 1, so the outCapacity will be 0?... Don't understand this at all.

MainUnicode.cpp
```diff
@@ -166,16 +208,16 @@ class ConvertingString
 	SQLCHAR * convUnicodeToString( SQLWCHAR *wcString, int length )
```
irodushka, line 166: I think all this method should be rewritten. I strongly suspect that the use of `wcstombs` here is questionable. 
std::c16rtomb or iconv...

irodushka:
@fdcastel I think it's a time to separate the ASAN/VALGRIND stuff from the unicode bugfix. I thought about this before and decided that the fix was so small and trivial that there was no point in creating a dedicated PR, but now... phew...

fdcastel:
> that the fix was so small and trivial that there was no point in creating a dedicated PR, but now... phew...

Same here. Fully agreed. 

I’ll roll back the Unicode fixes and move them into separate tasks for the [next version plan](https://github.com/FirebirdSQL/firebird-odbc-driver/issues/287).

irodushka:
> > that the fix was so small and trivial that there was no point in creating a dedicated PR, but now... phew...
> 
> Same here. Fully agreed.
> 
> I’ll roll back the Unicode fixes and move them into separate tasks for the [next version plan](https://github.com/FirebirdSQL/firebird-odbc-driver/issues/287).

Okay, thank you. Sorry if I'm being too boring))
FYI - this file from the unixODBC source, look at `ansi_to_unicode_copy()` and `unicode_to_ansi_copy()` functions.
[__info.c](https://github.com/user-attachments/files/26858873/__info.c)

P.S. We cannot ~~borrow~~ use them as is, but can investigate the approach)

fdcastel:
> Okay, thank you. Sorry if I'm being too boring)) 

Haha, you’re not at all. Being critical actually really helps here! 😉 

> P.S. We cannot ~borrow~ use them as is, but can investigate the approach)

Yep! Got it! Thanks for the pointers!

I'll start to work on this soon.

fdcastel:
All done!

Rolled back the Unicode changes from this PR so #289 stays strictly scoped to ASAN/Valgrind.

The deferred Unicode follow-ups have been moved into a new Tier 9.1 in issue #287, to be handled after Tier 9.

fdcastel:
@irodushka if that works for you, once this is merged I’ll start working on PRs for the “Tier 1” and “Tier 2” sections from #287.

irodushka:
@fdcastel 

Hmmm. That means - all the ASAN builds will fail, and the driver itself continues to be "toxic" - this memory corruption ASAN warning is NOT a harmless thing, you understand, and may lead to a real crash.

Can you please remind me - those unicode fixes&improvements that were planned from the very begining, did they include the fix we're talking about now, or is this a new issue?

MainUnicode.cpp
```diff
@@ -85,7 +85,7 @@ class ConvertingString
 			if ( length == SQL_NTS )
 				lengthString = 0;
 			else if ( retCountOfBytes )
-				lengthString = length / sizeof(wchar_t);
+				lengthString = length / sizeof(SQLWCHAR);
```
irodushka: And please! Have you forgotten that this change causes a stack smash?)) Simply because the the lengthString becomes _greater_, and the `mbstowcs` call below begins to break the client's stack, overflowing the `unicodeString`, passed by the pointer to the ODBC call?

This undoubtedly right change CAN NOT be done without a deep refactoring of the unicode routins. I thought I've explaned it clearly in https://github.com/FirebirdSQL/firebird-odbc-driver/pull/289#issuecomment-4263027888

irodushka (changes requested):

CMakePresets.json
```diff
@@ -30,6 +30,22 @@
             "cacheVariables": {
                 "CMAKE_BUILD_TYPE": "Debug"
             }
+        },
+        {
+            "name": "asan",
+            "displayName": "Debug + AddressSanitizer",
+            "inherits": "debug",
+            "cacheVariables": {
+                "BUILD_WITH_ASAN": "ON"
```
irodushka: Should also add `"BUILD_WITH_VALGRIND":"OFF"` here. 
Otherwise we can get the both options set (ASAN from here and VALGRIND from the cache) and the consequent cmake fatal error ("BUILD_WITH_ASAN and BUILD_WITH_VALGRIND are mutually exclusive" and so on).

CMakePresets.json
```diff
@@ -30,6 +30,22 @@
...
+            }
+        },
+        {
+            "name": "valgrind",
+            "displayName": "Debug + Valgrind",
+            "inherits": "debug",
+            "cacheVariables": {
+                "BUILD_WITH_VALGRIND": "ON"
```
irodushka: Should also add "BUILD_WITH_ASAN":"OFF" here.

irodushka (changes requested):

CMakePresets.json
irodushka, line 8: And maybe it's a good idea to add both the
```
"BUILD_WITH_VALGRIND":"OFF",
"BUILD_WITH_ASAN":"OFF"
```
to all standard build configurations (except the dedicated for ASAN/Valgrind).

fdcastel:
> this memory corruption ASAN warning is NOT a harmless thing, you understand, and may lead to a real crash.

I understand and agree. However, the driver is currently running fine as is.

Yes, we’ve identified a possible bug, but we’re not receiving any complaints. Given that, I’d suggest leaving it as-is for now (see below).


> Can you please remind me - those unicode fixes&improvements that were planned from the very begining, did they include the fix we're talking about now, or is this a new issue?

I’ve moved it to Phase ("Tier") 9.1. Unicode was completely overhauled in v4, so by the time we reach Phase 9, this issue may no longer even exist.



_Working on your last review now._

MainUnicode.cpp
fdcastel: You were right — this change alone breaks the symmetric `mbstowcs((wchar_t*)unicodeString, ..., lengthString)` call in the destructor, which then writes `lengthString * sizeof(wchar_t)` bytes into the caller's SQLWCHAR buffer and smashes the stack. Fixing the capacity without also rewriting the write-back path is not viable. Reverted in f20e6fc; the `ConvertingString` / `mbstowcs` refactor has been tracked as Tier 9.1 in #287. PR #289 is now strictly scoped to the ASAN/Valgrind CI plumbing plus the safe internal-buffer allocation fix in 832d8e7.

CMakePresets.json
fdcastel: Done in 9daab27 — the `asan` preset now explicitly sets `BUILD_WITH_VALGRIND: OFF`.

CMakePresets.json
fdcastel: Done in 9daab27 — the `valgrind` preset now explicitly sets `BUILD_WITH_ASAN: OFF`.

CMakePresets.json
fdcastel: Good idea — done in 8258a27. `default`, `release`, and `debug` presets now all pin `BUILD_WITH_ASAN: OFF` and `BUILD_WITH_VALGRIND: OFF`, so switching back from a sanitizer preset can't carry a stale flag forward via the CMake cache.

MainUnicode.cpp
irodushka: Right. But it's insufficient to close this thread, please revert the changes) return the line 88 back to the original
`lengthString = length / sizeof(wchar_t);`

irodushka (changes requested):

irodushka:
> However, the driver is currently running fine as is.
> 
> Yes, we’ve identified a possible bug, but we’re not receiving any complaints.

:))) This is because 99.9% of customers use the ODBC driver with Windows. And 99.9% of those 99.9% use it with MS Office - Word, Excel & PowerBI.

Okay. Just another reason to make the rest of the work fast (the main brake of the process has said:)

MainUnicode.cpp
irodushka: > Reverted in [f20e6fc](https://github.com/FirebirdSQL/firebird-odbc-driver/commit/f20e6fc36f3dcdca3b5dc93ec600dc6446160c18);

Cannot see where the line 88 is reverted in ^^^

irodushka (changes requested):

fdcastel:
Please wait. Trying one more thing...

irodushka:
> Please wait. Trying one more thing...

Okay, I'm in no hurry)

Just one thing to say - I looked into PR #283 and, frankly, I don't like what I see in the MainUnicode.cpp and in the reworked `class ConvertingString` in particular. Overcomplicated, all that redundant(?) logic with multi-bytes (surrogates etc)... Do I understand right you gonna refactor it after our last investigations?

fdcastel:
> Okay, I'm in no hurry)

🤣 
 
> I looked into PR #283 and, frankly, I don't like what I see in the MainUnicode.cpp and in the reworked `class ConvertingString` in particular. Overcomplicated, all that redundant(?) logic with multi-bytes (surrogates etc)... Do I understand right you gonna refactor it after our last investigations?

That was just the first attempt 😉 -- no worries! 

SURE, We will fix all that. `ConvertingString` as it is today should be [killed with fire](https://knowyourmeme.com/memes/kill-it-with-fire).

fdcastel:
Reverted back to `length / sizeof(wchar_t)` -- the destructor's `mbstowcs` writeback is now safe again (no more stack smash).

fdcastel:
P.S: Added a minimum-size floor in `Alloc()`.

internal `byteString` is now at least 8 bytes, so `OdbcError::sqlGetDiagRec`'s `strcpy("HY000\0")` no longer overflows even when `lengthString=3`.

This closes the original heap-buffer-overflow without touching the `mbstowcs` bound.

MainUnicode.cpp
```diff
@@ -219,8 +219,16 @@ class ConvertingString
...
-				byteString = new SQLCHAR[ lengthString + 2 ];
-				memset(byteString, 0, lengthString + 2); 
+				// Floor the internal buffer at 8 bytes so that callers which pass a
+				// small SQLWCHAR output buffer (e.g. SQLGetDiagRecW with a 12-byte
+				// SQL state, yielding lengthString=3 on Linux where sizeof(wchar_t)=4)
+				// still have room for the 6-byte SQL state ("HY000\0") that
+				// OdbcError::sqlGetDiagRec strcpy's into this buffer. Keeping
+				// lengthString itself unchanged preserves the mbstowcs writeback
+				// bound and avoids smashing the caller's stack buffer.
+				const size_t bufSize = (lengthString + 2 < 8) ? 8 : (size_t)lengthString + 2;
```
irodushka: Hmmm. Maybe not doing that?.. It's smth we call in Russia "a crutch")
Don`t forget that the 5-digit state is not the only scenario, the other is with the full error description - see SQLGetDiagRecW():
```
ConvertingString<> MessageText( bufferLength, messageText, textLength, isByte );
``` 

And the other thing - even if we avoid the strcpy overflow, the data in the unicodeString (after mbstowcs() call) will contain UTF32 characters, that is not expected by the client side (the client expects UCS-2 or fixed UTF-16). So, it's incorrect in any case.

fdcastel: I dropped the crutch and pushed `bea6403`. Please take a look at it, now.

My floor-at-8 was trying to close the overflow without admitting that the writeback itself was wrong, and you were right to push back. 👍🏻

MainUnicode.cpp
```diff
@@ -135,13 +150,33 @@ class ConvertingString
...
+						}
+						len = i;
+					}
 #endif
 				}
 
 				if ( len > 0 )
 				{
-					*(LPWSTR)(unicodeString + len) = L'\0';
+					// NUL-terminate in SQLWCHAR units. LPWSTR assignment of L'\0' writes
```
irodushka: Who will overrun what?

LPWSTR is a pointer to WCHAR - sqltypes.h:112
WCHAR is unsigned short - sqltypes.h:109. Yes, if you #define SQL_WCHART_CONVERT then it may be wchar_t, but the SQL_WCHART_CONVERT is NOT defined by default.
So, sizeof(*LPWSTR) == 2. Not 4, not 42, not 100500, but just 2.

And anyway - see sqltypes.h:467 - `typedef WCHAR         SQLWCHAR`! And, consequently, LPWSTR == SQLWCHAR*. 

Do you suppose that if we assign L'\0', that is obviously int32_t (4 bytes) size, to the 2 byte variable (SQLWCHAR type) we get an overrun? If you do, you're totally wrong. We'll write exactly the number of bytes that our variable can hold - 2 bytes. The remaining (higher) 2 bytes will be simply discarded. No overrun, no overflow - this is called "narrowing". The narrowing may result in the data loss (and you get a compiler warning in that case), but while we do it with the pure zero (L'\0' == 0x00000000) it's nothing to loose.

Yes, you can leave this change - I mean `unicodeString[len] = 0;` - it's more compact and understandable. But please do not write such a strange comments.

irodushka (changes requested):

irodushka:
@fdcastel 

I really appreciate your work, but I think we're taking a wrong turn.

1. I remember we decided to separate the ASAN/VALGRING stuff from the unicode fixes. But now you are mixing it all up again. 
2. When I wrote `That means - all the ASAN builds will fail, and the driver itself continues to be "toxic"...` I meant that it's better to fix the unicode issue asap, not postponing it until some 9.* phase, but a separation is still mandatory. It can be two subsequent PRs, for instance.
3. When I have to choose between the development speed and code quality, I always (almost always:) choose the latter. You know the saying "Measure twice, cut once"?) We have a Russian very close equivalent, but with "measure **seven** times", not twice:)

irodushka:
So. What about following the plan written below?)

1. No need to rush! Very important bullet!
2. Clean up this PR - leave only ASAN/Valgrind issues. And merge it. ASAN build will fail - ignore it.
3. Open the separated PR for unicode issues.
4. Develop a bundle of tests for widechar ODBC calls. These tests must cover the both usage scenarios of ConvertingString class (both of the ctor calls). I suppose that initially all those tests will fail on Linux.
5. Then I myself will try to implement the correct logic. Having the tests is critical here.
6. After that you'll merge these changes back into your branch.
7. Everyone is happy and ~~gay~~ no you cannot say ~~"gay"~~ in that meaning no no no.

fdcastel:
Updated #287: Moved Tasks from Tier 9.1 to Tier 1b.

I’ll be working on the split PRs over the next few hours.

irodushka:
> Updated #287: Moved Tasks from Tier 9.1 to Tier 1b.
> 
> I’ll be working on the split PRs over the next few hours.

The primary task (after the asan/valgrind PR cleaning) is to develop the widechar odbc tests. Okay?

fdcastel:
> The primary task (after the asan/valgrind PR cleaning) is to develop the widechar odbc tests. Okay?

Yep. Loud and clear! 😉

fdcastel:
@irodushka — split done per your plan:

- **#289** (this PR) trimmed to ASAN/Valgrind CI only. `MainUnicode.cpp` reverted to master. ASAN job is `continue-on-error` while the Unicode fix is pending (Valgrind stays strict). Ready to merge.
- **#290** — widechar ODBC call tests covering both `ConvertingString` ctor paths (output: `SQLGetDiagRecW`/`SQLErrorW`; input: `SQLExecDirectW`/`SQLPrepareW`). Linux runs skip with a pointer back here; Windows runs them for real.
- **#291** — *draft* reference implementation of the Linux Unicode fix against #290's tests. Yours to approve, rewrite on top of, or close-and-replace — whichever fits step 5 of your plan.

Suggested order: merge #289 → review #290 → take over #291.

fdcastel:
Just a reminder: these are items that were intentionally left out of scope for this plan/PR.

* The `connection->MbsToWcs` / `WcsToMbs` Linux branch in `MainUnicode.cpp` (around lines 129 and 195) is still incorrect in the same way, but it is not exercised by the current test surface.
* True `iconv` / UTF-8 correctness for non-ASCII input.
* Redesign of `ConvertingString` (destructor-as-converter, silent error handling, etc.).

All of these belong in #287 (Tier 1b). After #291 has landed.

fdcastel:
P.S.: As expected, ASAN is failing now. It will be green after #291, too. 😉

irodushka:
I think we should remove the `Draft` tag, disable (temporarily) the ASAN build&test and merge this PR to the master branch.
Are you ok with it?

fdcastel:
Agreed. Done in 98115a1:

- ASAN matrix entry commented out (with a pointer back to #291 / #287 Tier 1b for re-enabling).
- `continue-on-error` hack removed — no longer needed.
- Valgrind stays strict.
- PR marked ready for review.

Leaving the merge to you 😉.

fdcastel:
I need a beer! 🤣 

P.S; @irodushka Thank you very much for your patience and guidance. Much appreciated! 🤗

irodushka:
> I need a beer! 🤣

Come to Moscow))

fdcastel:
> Come to Moscow))

Second invite in less than a month. Adding it to my ever-growing “dangerously tempting ideas” list. 😅

PR #290, opened 2026-04-20
Add widechar ODBC call tests (driving Unicode fix)

fdcastel:
## Summary

Adds a dedicated test file `tests/test_wide_errors.cpp` that exercises both `ConvertingString` constructor scenarios in `MainUnicode.cpp`:

- **Output path** — `SQLGetDiagRecW`, `SQLErrorW` (uses `ConvertingString(length, sqlState)` to widen the error text back to the caller's SQLWCHAR buffer).
- **Input path** — `SQLExecDirectW`, `SQLPrepareW` (uses `ConvertingString(connection, wcString, length)` / `convUnicodeToString` to narrow the caller's SQLWCHAR query into bytes).

## Why this PR exists (and why it's separate from #289)

Per @irodushka's plan in https://github.com/FirebirdSQL/firebird-odbc-driver/pull/289#issuecomment-4279111183:

1. ASAN/Valgrind CI stays strictly scoped to #289.
2. This PR provides the regression coverage for the known Linux widechar bugs (truncation to one SQLWCHAR on read-back, stack smash on small buffers).
3. A follow-up PR will ship the `MainUnicode.cpp` rewrite and un-skip these tests on Linux.

## Linux behaviour (today)

Every test in the new file starts with:

```cpp
SKIP_ON_LINUX_WIDECHAR();
```

which is a `GTEST_SKIP` with a pointer to issue #287 **Tier 1b**. The pattern matches existing per-platform skips in [tests/test_wchar.cpp:111](tests/test_wchar.cpp#L111) and [tests/test_errors.cpp](tests/test_errors.cpp). This keeps Linux CI green; Windows runs them for real and they must pass. Removing the skip is the visible diff in the follow-up Unicode-fix PR.

## Test plan

- [ ] CI green on Linux (tests show as SKIPPED)
- [ ] CI green on Windows (tests run and pass)
- [ ] Locally remove the `SKIP_ON_LINUX_WIDECHAR()` guards on Linux and confirm the tests fail the way we expect (truncation / overrun) — that's what makes them useful as driving tests for the follow-up PR.

## Related

- Issue #288, PR #289 (ASAN/Valgrind CI).
- Issue #287 Tier 1b (Unicode rewrite — consumed by the follow-up PR).

PR #291, opened 2026-04-20
Fix SQLWCHAR conversion paths on Linux (draft)

fdcastel:
## Summary

Draft reference implementation against the widechar test suite in #290. Fixes the `MainUnicode.cpp` Linux branches that assumed `sizeof(SQLWCHAR) == sizeof(wchar_t)`.

> **Note:** this branch is layered on top of #290, so the diff here includes both the tests (from #290) and the Unicode fix. Review the commits individually, or wait for #290 to merge and this will rebase cleanly to show only the Unicode fix.

## What this PR does

- **`ConvertingString` destructor** (output path, used by `SQLGetDiagRecW` / `SQLErrorW`): Linux branch now does a byte-level widening loop (`SQLCHAR` → `SQLWCHAR`) instead of `mbstowcs((wchar_t*)...)` — the latter wrote UTF-32 into a UTF-16 buffer on Linux, truncating `HY000` to `H` and smashing the stack on tight output buffers.
- **`convUnicodeToString`** (input path, used by `SQLExecDirectW` / `SQLPrepareW`): Linux branch does the inverse byte-narrowing loop instead of `wcstombs`, and uses a local `sqlwcharLen` helper instead of `wcslen((wchar_t*)...)`.
- **`lengthString = length / sizeof(SQLWCHAR)`**: correct character count in the length-specified ctor.
- **NUL-terminate in SQLWCHAR units.** In `convUnicodeToString`, `*ptEndWC = L'\0'` through a `wchar_t*` was a genuine 4-byte write into a 2-byte `SQLWCHAR` slot on Linux and is the real fix. The sibling `*(LPWSTR)...` in the destructor is a 2-byte narrowing store — not an overrun, per @irodushka in #289 (discussion_r3109004999) — so the replacement there is a style simplification, not a fix.

Mirrors unixODBC's own `ansi_to_unicode_copy` / `unicode_to_ansi_copy` approach.

## Scope / Out of scope

- Does **not** close Tier 1b entirely — the `connection->MbsToWcs` / `WcsToMbs` branch and non-ASCII (iconv) handling remain open.
- This PR also removes the Linux-only `GTEST_SKIP` guards added in #290, so the test suite runs on every platform.

## Context

Per @irodushka's plan in #289 comment 4279111183. Opened as **draft** so either:
(a) @irodushka approves and it lands,
(b) @irodushka pushes his own implementation on top, or
(c) @irodushka closes it and opens his own, using #290's tests as the acceptance criteria.

## Test plan

- [ ] CI run green on all platforms including ASAN strict
- [ ] Windows: widechar tests from #290 still pass
- [ ] Linux: widechar tests from #290 now pass (skips removed)
- [ ] Valgrind: no new leaks

### DIFF
--- a/MainUnicode.cpp
+++ b/MainUnicode.cpp
@@ -38,6 +38,21 @@
 extern FILE	*logFile;
 using namespace OdbcJdbcLibrary;
 
+#ifndef _WINDOWS
+// SQLWCHAR-aware length (in SQLWCHAR units), safe on Linux where
+// sizeof(wchar_t) != sizeof(SQLWCHAR). Do NOT use wcslen() on SQLWCHAR
+// data on Linux — it reads two SQLWCHARs per wchar_t and runs off the end.
+static size_t sqlwcharLen( const SQLWCHAR *s )
+{
+	size_t n = 0;
+	if ( !s )
+		return 0;
+	while ( s[n] )
+		++n;
+	return n;
+}
+#endif
+
 #ifdef _WINDOWS
 extern UINT codePage; // from Main.cpp
 #endif
@@ -85,7 +100,7 @@ public:
 			if ( length == SQL_NTS )
 				lengthString = 0;
 			else if ( retCountOfBytes )
-				lengthString = length / sizeof(wchar_t);
+				lengthString = length / sizeof(SQLWCHAR);
 			else
 				lengthString = length;
 		}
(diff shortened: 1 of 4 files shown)


irodushka:
@fdcastel 

>NUL-terminate in SQLWCHAR units (no LPWSTR 4-byte write past the end on Linux).

Please read again this explanation https://github.com/FirebirdSQL/firebird-odbc-driver/pull/289#discussion_r3109004999
I tried to explain that the changes you made were not a bug fix, but an optional code styling. That the C/C++ code like
```
int64_t x = 0;
int32_t y = x;
```
or
```
int64_t x = 0;
int32_t y;
int32_t *py = &y;
*py = x;
```
contains no bug.
Well, maybe you think I'm wrong? Let's discuss it.

irodushka:
ok!

fdcastel:
@irodushka Agreed on the destructor case — `*(LPWSTR)ptr = L'\0'` is a 2-byte narrowing store on unixODBC (`LPWSTR == SQLWCHAR* == unsigned short*`), not an overrun. Already corrected in 0d9bce9: the overrun comment was dropped and `unicodeString[len] = 0;` was kept only as the style simplification you endorsed. The PR description still overclaimed it as a bug fix — fixing that now.

One distinction worth separating on the sibling case in `convUnicodeToString`: pre-PR code declared `wchar_t *ptEndWC` and did `*ptEndWC = L'\0';`. There the pointee type is already `wchar_t`, so no narrowing happens — the store is `sizeof(wchar_t) == 4` bytes on Linux into a 2-byte `SQLWCHAR` slot. That one is a genuine 4-byte overrun and the fix stays. Let me know if you read that one differently.

fdcastel:
PR description updated to reflect the distinction.

MainUnicode.cpp
```diff
@@ -170,12 +202,18 @@ class ConvertingString
...
 		if ( length == SQL_NTS )
+#ifdef _WINDOWS
 			length = (int)wcslen( (const wchar_t*)wcString );
-		else if ( wcString[length] != L'\0' )
+#else
+			length = (int)sqlwcharLen( wcString );
+#endif
+		else if ( wcString[length] != 0 )
 		{
 			ptEndWC = (wchar_t*)&wcString[length];
```
irodushka, line 212:
```suggestion
			ptEndWC = wcString + length;
```

MainUnicode.cpp
```diff
@@ -170,12 +202,18 @@ class ConvertingString
 		wchar_t saveWC;
```
irodushka, line 202:
```suggestion
		SQLWCHAR saveWC;
```

MainUnicode.cpp
irodushka, line 201:
```suggestion
		SQLWCHAR *ptEndWC = NULL;
```

irodushka (changes requested):

irodushka:
Yep. This place
```
			ptEndWC = (wchar_t*)&wcString[length];
			saveWC = *ptEndWC;
			*ptEndWC = L'\0';
```
really looks like overrun. Frankly, two overruns - 1) `*ptEndWC = L'\0';` - you've fixed it 2) `saveWC = *ptEndWC;` - out-of-bound read, not fixed yet)) A read is not so deadly as a write, but it can lead to UB or even crash.

and in the second place, at the end of the function:
```
		if ( ptEndWC )
			*ptEndWC = saveWC;
```
this's the third real write overrun)) unfixed)

I think it's easier to rewrite this than to try to fix all those issues...

fdcastel:
image

PR #293, opened 2026-04-22
tests: cover sibling numeric→VARCHAR conversions

fdcastel:
## Summary

Builds on top of #292.  That PR fixed the `conv*ToString` helpers so numeric C-types bound to a Firebird VARCHAR parameter no longer corrupt the VARYING length prefix, but it only added test coverage for the `convLongToString` path (the single scenario reported in duckdb/odbc-scanner#161).  A regression in any of the sibling helpers — `convShortToString`, `convBigintToString`, `convFloatToString`, `convDoubleToString` — would slip through CI.

This PR adds one compact, loop-based test per sibling helper:

- `SShortToVarcharLoopDml`    — `SQL_C_SSHORT` → `convShortToString`
- `SBigintToVarcharLoopDml`   — `SQL_C_SBIGINT` → `convBigintToString`
- `FloatToVarcharLoopDml`     — `SQL_C_FLOAT` → `convFloatToString`
- `DoubleToVarcharLoopDml`    — `SQL_C_DOUBLE` → `convDoubleToString`

Each binds the relevant C-type to a VARCHAR column, executes the same prepared `INSERT INTO ... VALUES (?, ?)` across a small set of values chosen to exercise 1-, 2- and 3-digit widths plus a negative, and verifies:
- every row committed,
- no stored value contains an embedded `\0` byte,
- `CAST(val AS <original-type>)` round-trips for every boundary value.

DML (no stored procedure) is used so the tests run everywhere the existing parameterized-statement coverage runs; the FB 6 master parameterized-query regression is skipped via `SKIP_ON_FIREBIRD6()` — same rationale as #292's Issue161 tests.

## Scope notes

- `SQL_C_TYPE_DATE` / `SQL_C_TYPE_TIMESTAMP` → VARCHAR are **intentionally omitted**.  The driver routes those C-types through `convDateToString` / `convDateTimeToString`, which expect Firebird's internal `ISC_DATE` / `ISC_TIMESTAMP` encoding (a 32-bit int and a QUAD), NOT the ODBC-side `SQL_DATE_STRUCT` / `SQL_TIMESTAMP_STRUCT` the application actually binds.  The resulting VARCHAR is garbage (year 2043, etc.) independent of the length-prefix fix, so this is a separate pre-existing bug that should get its own fix and its own tests.
- `SQL_C_STINYINT` / `SQL_C_TYPE_TIME` similarly untested — low-value coverage.

## Dependency

Depends on #292.  Without that PR's `conv*ToString` length-prefix fix, every sibling test here exhibits the same silent data loss the SLONG test did (verified locally against the old v3.0.1.21 driver — 11/6 rows stored instead of the expected row count).  GitHub will show this PR's diff cleanly once #292 merges.

## Test plan

- [x] All 10 matrix jobs green: Windows x64 / x86 / ARM64 and Linux x64 / ARM64 on both Firebird 5.0.3 and Firebird master, plus Valgrind on Linux 5.0.3.

### DIFF
--- a/IscDbc/Connection.h
+++ b/IscDbc/Connection.h
@@ -524,6 +524,7 @@ public:
 	//virtual void		setSqlSubType ( short subtype ) = 0;
 	virtual void		setSqlLen ( short len ) = 0;
 	virtual short		getSqlMultiple () = 0;
+	virtual bool		isSqlVarying () = 0;
 
 	virtual char *		getSqlData() = 0;
 	virtual short *		getSqlInd() = 0;
--- a/IscDbc/IscHeadSqlVar.h
+++ b/IscDbc/IscHeadSqlVar.h
@@ -102,6 +102,7 @@ public:
 	inline short	getSqlMultiple () { return sqlMultiple; }
 	inline char *	getSqlData() { return sqlvar->sqldata; }
 	inline short *	getSqlInd() { return sqlvar->sqlind; }
+	inline bool		isSqlVarying() { return sqlvar->sqltype == SQL_VARYING; }
 
 	// not used
 	//void		setSqlInd( short *ind ) { sqlvar->sqlind = ind; }
--- a/IscDbc/Sqlda.cpp
+++ b/IscDbc/Sqlda.cpp
@@ -759,7 +759,20 @@ const char* Sqlda::getColumnName(int index)
 
 int Sqlda::getPrecision(int index)
 {
-	CAttrSqlVar *var = Var(index);
+	CAttrSqlVar *curVar = Var(index);
+	// Issue #161: for INPUT parameters the conversion code mutates sqlvar
+	// (via setTypeText/setSqlLen).  Reading precision from the current
+	// sqlvar makes `record->length` shrink after the first row was bound,
+	// which then causes subsequent numeric→VARCHAR conversions to write
+	// only the first character of multi-digit values (silent data loss).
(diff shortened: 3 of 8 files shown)


irodushka:
And with this?.. I don't understand

fdcastel:
Sorry — same delay as #292.  Posting an update now that both PRs are rebased on current master.

This PR adds sibling-numeric test coverage (`SShort`, `SBigint`, `Float`, `Double` → VARCHAR) on top of the `SLong` coverage in #292.  The driver-side fix lives in #292; once that lands, this PR reduces to a clean tests-only diff.

Rebased on master `b64c011` (post #294's `SQL_DBMS_VER` fix and #297's `SQLCopyDesc` SEGV fix).  All 10 matrix jobs green on the latest run ([25896361594](https://github.com/fdcastel/firebird-odbc-driver/actions/runs/25896361594)).  No conflicts.

The branch currently bundles #292's driver commits because the sibling tests can't pass without that fix; they'll auto-resolve via patch-id matching when #292 merges.  Happy to restructure differently if you'd prefer — e.g. base this branch directly on #292's head, or strip the bundled commits and accept that CI here goes red until #292 lands.

fdcastel:
Folded into #292 as its third commit (the four sibling loop tests), so there is no stacked PR showing foreign commits any more. Closing this one.

Issue #169, opened 2014-11-03
Character set conversion errors in SQLGetData when reading BLOB SUB_TYPE 1 [ODBC180]


firebird-automations:
Submitted by: Jojakim Stahl (jstahl)

Attachments:
[bug-180.patch](https://github.com/FirebirdSQL/jira-attachments/raw/main/ODBC/ODBC-180/12610_bug-180.patch)

When reading a charachter type BLOB as LONG\(W\)VARCHAR, no/wrong charset conversion is performed\.
When char BLOB is bound to SQL\_C\_WCHAR, mbstowcs conversion cannot be applied to each chuck returned from the DB seperately, but must be done in advance for the complete BLOB\. Imagine i\.e\. a 3\-byte UTF8 sequence cut split over 2 chunks\. Impossible to convert correctly, therefore need to convert all at once in advance\.

Commits: FirebirdSQL/firebird-odbc-driver@9fca7df5b63b52370aa5387a7fa739804f8fa45e FirebirdSQL/firebird-odbc-driver@d94a113365a56c244d7f66d54853f8e53a4e073a

PR #296, opened 2026-04-27
Fix SQL_C_GUID parameter binding sending corrupted data on the wire (#295)

### LINKED ISSUE #295: SQL_C_GUID parameter binding sends corrupted data on the wire (Phase 8 / T5-5)
## Summary

When an ODBC application calls `SQLBindParameter` with C type `SQL_C_GUID` and SQL type `SQL_GUID` and provides a 16-byte UUID buffer, the driver **accepts the call without error** but **does not convert the C-side 16-byte GUID into Firebird's `BINARY(16)` / `CHAR(16) CHARACTER SET OCTETS` wire format**. The data observed on the Firebird side is corrupted (and the corruption pattern depends on the inferred SQL parameter type).

This is captured in [#287 Tier 5 / T5-5](https://github.com/FirebirdSQL/firebird-odbc-driver/issues/287) (\"SQL_GUID type mapping\") as ❌ OPEN. Filing this as a concrete reproducer with empirical observations so the eventual T5-5 fix has a clear acceptance test.

## Versions tested

- Driver: **firebird-odbc 3.0.1.21** (current Chocolatey release) and **v3.5.0-rc1** sources (the new Phase 8 GUID test file `tests/test_guid_and_binary.cpp` is present, but every test is `GTEST_SKIP() << \"Requires Phase 8: SQL_GUID type mapping and FB4+ types (not yet merged)\"`).
- Firebird: **5.0.3.1** (Windows Server 2025).
- Application: [duckdb/odbc-scanner](https://github.com/duckdb/odbc-scanner), which binds DuckDB `::UUID` parameters via `SQLBindParameter(SQL_C_GUID, SQL_GUID, len=16, ptr=16-byte-GUID, ...)` (see [src/types/uuid_type.cpp lines 33-44](https://github.com/duckdb/odbc-scanner/blob/main/src/types/uuid_type.cpp#L33-L44)).

## Reproducer

The failing CI run is on a paused branch on my fork: [`fdcastel/odbc-scanner:firebird-uuid-test`](https://github.com/fdcastel/odbc-scanner/tree/firebird-uuid-test). Two CI runs document the behavior:

### Reproducer 1 — `UUID_TO_CHAR(?)` (parameter inferred by Firebird as `BINARY(16)`)

```sql
SELECT UUID_TO_CHAR(?) FROM rdb$database
-- bound parameter: SQL_C_GUID, 16 bytes, value = UUID '0306C176-E401-11F0-A1AC-E86A6416A6F5'
```

- **Expected** (clean round-trip): `0306C176-E401-11F0-A1AC-E86A6416A6F5`
- **Actual** (CI [run 25007127097, attempt 2](https://github.com/fdcastel/odbc-scanner/actions/runs/25007127097)): `30333036-4331-3736-2D45-3430312D3131`

The actual 16 bytes Firebird stored are `30 33 30 36 43 31 37 36 2D 45 34 30 31 2D 31 31` — i.e., the **ASCII characters** of the first 16 chars of the canonical-form UUID string. The driver appears to convert the SQL_C_GUID buffer to a 36-char canonical string and then truncate to 16 bytes when binding to a BINARY(16) parameter slot.

### Reproducer 2 — `CHAR_TO_UUID(?)` (parameter inferred by Firebird as `VARCHAR`)

```sql
SELECT UUID_TO_CHAR(CHAR_TO_UUID(?)) FROM rdb$database
-- same SQL_C_GUID parameter as above
```

- **Expected**: `0306C176-E401-11F0-A1AC-E86A6416A6F5`
- **Actual** (CI [run 25007671590](https://github.com/fdcastel/odbc-scanner/actions/runs/25007671590)):

  ```
  HY000(-833): [ODBC Firebird Driver][Firebird]expression evaluation not supported
  -Human readable UUID argument for CHAR_TO_UUID must have hex digit at position 2 instead of \"\"
  ```

  CHAR_TO_UUID receives an essentially empty string. Pattern is consistent with the SQL_C_GUID buffer being interpreted as a wide-char (UTF-16) C string by a narrow-char path — the leading `0x00` of `0x00 0x30 0x00 0x33 ...` terminates the string at position 1.

## Source code analysis (v3.5.0-rc1)

[`OdbcConvert.cpp:1003`](https://github.com/FirebirdSQL/firebird-odbc-driver/blob/v3.5.0-rc1/OdbcConvert.cpp#L1003) — the only place `SQL_C_GUID` appears as a *source* `conciseType`:

```cpp
case SQL_C_GUID:
    switch(to->conciseType)
    {
    case SQL_C_CHAR:
        return &OdbcConvert::convGuidToString;
    case SQL_C_WCHAR:
        return &OdbcConvert::convGuidToStringW;
    default:
        return &OdbcConvert::notYetImplemented;
    }
    break;
```

So `SQL_C_GUID → SQL_C_CHAR/WCHAR` is implemented (the *output* direction — fetching a GUID column as a string), but **`SQL_C_GUID → SQL_C_BINARY` and the parameter-binding direction generally is not implemented**.

[`OdbcStatement.cpp:664`](https://github.com/FirebirdSQL/firebird-odbc-driver/blob/v3.5.0-rc1/OdbcStatement.cpp#L664) and [`:2218`](https://github.com/FirebirdSQL/firebird-odbc-driver/blob/v3.5.0-rc1/OdbcStatement.cpp#L2218) — `SQL_C_GUID` appears in the accept-list of valid C types in `SQLBindCol` / `SQLBindParameter` (`case SQL_C_GUID: break;`), so the driver does not reject the call. But there is no corresponding wire-format conversion when the parameter is later sent.

[`tests/test_guid_and_binary.cpp`](https://github.com/FirebirdSQL/firebird-odbc-driver/blob/v3.5.0-rc1/tests/test_guid_and_binary.cpp) — every one of the eight SQL_GUID tests (`InsertAndRetrieveUuidBinary`, `UuidToCharReturnsValidFormat`, `CharToUuidRoundtrip`, `RetrieveAsSqlGuidStruct`, etc.) starts with:

```cpp
GTEST_SKIP() << \"Requires Phase 8: SQL_GUID type mapping and FB4+ types (not yet merged)\";
```

confirming the maintainers consider full SQL_GUID support a future (Phase 8) item. None of those tests currently cover the **parameter-binding direction** specifically — they all use `INSERT ... VALUES (GEN_UUID())` or `INSERT ... VALUES (CHAR_TO_UUID('literal'))` to populate columns server-side, then fetch.

## Expected behavior

`SQLBindParameter(stmt, n, SQL_PARAM_INPUT, SQL_C_GUID, SQL_GUID, 16, 0, ptr, 16, &len)` followed by `SQLExecute` should send the 16-byte GUID buffer over the wire so Firebird sees:

- For a `BINARY(16)` / `CHAR(16) CHARACTER SET OCTETS` parameter slot: those exact 16 bytes (in Firebird's native byte order — note that `SQL_GUID` Data1/Data2/Data3 are little-endian on x86, while `BINARY(16)` is byte-order-agnostic, so the conversion must explicitly arrange the 16 bytes per the canonical UUID layout).
- For a `VARCHAR`/`CHAR` parameter slot: the 36-char canonical-form UUID string, narrow-char encoded.

The fix maps to [#287 T5-5](https://github.com/FirebirdSQL/firebird-odbc-driver/issues/287) (\"add GUID conversion methods (GUID↔string, GUID↔binary)\"). Suggested acceptance test: un-skip `CharToUuidRoundtrip`, `InsertAndRetrieveUuidBinary`, and add a parallel test that uses `SQLBindParameter(SQL_C_GUID, ...)` to insert a known UUID (rather than `CHAR_TO_UUID('literal')`), then fetches it back and compares.

## Downstream

A duckdb/odbc-scanner [PR #169](https://github.com/duckdb/odbc-scanner/pull/169) follow-up adding a Firebird UUID round-trip test is paused on this issue. The branch [`fdcastel/odbc-scanner:firebird-uuid-test`](https://github.com/fdcastel/odbc-scanner/tree/firebird-uuid-test) has the test ready to merge once T5-5 lands.

fdcastel:
## What this fixes

When an ODBC application calls `SQLBindParameter(SQL_C_GUID, SQL_GUID, …, ptr, 16, &len)` and hands the driver a 16-byte UUID, the driver returns `SQL_SUCCESS` but Firebird never sees the 16 UUID bytes:

- **BINARY(16) / CHAR(16) CHARACTER SET OCTETS target** — Firebird stores the ASCII bytes of the first 16 chars of the canonical UUID string (`30 33 30 36 43 31 37 36 2D 45 34 30 31 2D 31 31` instead of the 16 raw UUID bytes).
- **Text target (e.g. `CHAR_TO_UUID(?)`) under CHARSET=UTF8** — Firebird returns `expression evaluation not supported — Human readable UUID argument for CHAR_TO_UUID must have hex digit at position 2 instead of ""`.

Reproducers, both from issue #295:

```sql
-- Reproducer A: parameter described as CHAR(16) CHARACTER SET OCTETS
SELECT UUID_TO_CHAR(?) FROM rdb$database
--   Expected: 'A0EEBC99-9C0B-4EF8-BB6D-6BB9BD380A11'
--   Actual:   '41304545-4243-3939-2D39-4330422D3445' (ASCII of the truncated string)

-- Reproducer B: parameter described as CHAR(36) in the connection charset
SELECT UUID_TO_CHAR(CHAR_TO_UUID(?)) FROM rdb$database
--   Expected: 'A0EEBC99-9C0B-4EF8-BB6D-6BB9BD380A11'
--   Actual:   HY000(-833) "must have hex digit at position 2 instead of ''"
```

Real-world impact: [duckdb/odbc-scanner](https://github.com/duckdb/odbc-scanner) binds DuckDB `::UUID` values exactly this way (see [`src/types/uuid_type.cpp` L33-44](https://github.com/duckdb/odbc-scanner/blob/main/src/types/uuid_type.cpp#L33-L44)). Their Firebird UUID round-trip test ([duckdb/odbc-scanner#169](https://github.com/duckdb/odbc-scanner/pull/169)) is parked on this.

Closes #295.

## Root cause

- **Reproducer A.** The driver stringified the GUID into the 16-byte OCTETS slot. On Windows `snprintf` is `_snprintf` (see `OdbcJdbc.h`), which truncates and returns -1; the legacy `len == -1` line then reported 16 bytes, so Firebird stored the ASCII of the first 16 chars.
- **Reproducer B.** With CHARSET=UTF8 Firebird describes the `?` in `CHAR_TO_UUID(?)` as CHAR(36) UTF8, 144 bytes. The driver maps UTF8 text to JDBC_WCHAR, so the dispatch picked `convGuidToStringW` and wrote UTF-16 onto the wire. Firebird saw `A`, NUL, `0`, NUL, … hence "hex digit at position 2". With CHARSET=NONE the slot is CHAR(36) ASCII and the old code passed on Windows by accident (36 chars fit `_snprintf` exactly, no terminator); C99 `snprintf` writes 35 chars plus NUL there.

## How the fix works

`OdbcConvert::getAdressFunction` switches on the target type first, like the neighbouring arms. For an input parameter described as CHAR/VARCHAR:

- **OCTETS charset** (Firebird's BINARY(n) / VARBINARY(n)): `convGuidToBinary` writes the 16 bytes in canonical UUID byte order (Data1/2/3 big-endian, Data4 as is). A slot shorter than 16 bytes is rejected with 22001.
- **Any other charset**: `convGuidToString` stages the 36-char canonical UUID in the DescRecord local buffer and points `sqldata` at it, the idiom `transferStringToAllowedType` uses. The wire buffer cannot be written directly: it may be VARYING (length prefix), shorter than 36 bytes, or a UTF8 text the driver maps to SQL_C_WCHAR. `setTypeText()` makes the slot SQL_TEXT and `Sqlda::checkAndRebuild()` rebuilds the message for the new length; Firebird reports the truncation itself when the slot is too small.
- **Any other target type** (INTEGER, DATE, BLOB): 07006, as before. ODBC defines no GUID conversion to those types.

`convGuidToString` also serves the application-side fetch path (`SQLGetData(…, SQL_C_CHAR, …)` from a column described as SQL_GUID, dormant until T5-5 / #287 lands); it branches on `to->isIndicatorSqlDa` only for the destination. `convGuidToStringW` is unchanged. Both Firebird-side writers use `ODBCCONVERT_CHECKNULL_SQLDA`, so a value bound after a NULL clears the null flag.

`IscHeadSqlVar::isBinary()` / `isVarBinary()` read the dedicated `sqlcharset` field to identify OCTETS slots.

## Tests

`GuidParamBindingTest` in `tests/test_guid_and_binary.cpp`:

- `BindGuidToCharOctets16`, `BindGuidToVarcharOctets16` — SQL_C_GUID into CHAR(16) / VARCHAR(16) OCTETS columns, read back via `UUID_TO_CHAR`.
- `BindGuidToUuidToCharRoundtrip`, `BindGuidToVarcharViaCharToUuid` — the two reproducers. The text one fails on the old code only under CHARSET=UTF8 (CI's primary run) or on Linux; the comment in the test says why.
- `BindGuidToIntegerParamIsRejected` (07006), `BindGuidToUndersizedOctetsIsRejected` (22001), `BindGuidToOversizedOctetsStoresRawBytes` (VARBINARY(32) receives 16 bytes), `BindGuidValueAfterNull` (a value bound after a NULL is sent).

The two VARCHAR OCTETS tests skip on the FB6 snapshot, which aborts parameterized OCTETS-varying statements server-side, the same limitation guarded elsewhere in the suite.

### DIFF
--- a/IscDbc/Connection.h
+++ b/IscDbc/Connection.h
@@ -524,6 +524,8 @@ public:
 	//virtual void		setSqlSubType ( short subtype ) = 0;
 	virtual void		setSqlLen ( short len ) = 0;
 	virtual short		getSqlMultiple () = 0;
+	virtual short		getSqlSubtype () = 0;
+	virtual short		getSqlLen () = 0;
 
 	virtual char *		getSqlData() = 0;
 	virtual short *		getSqlInd() = 0;
--- a/IscDbc/Connection.h
+++ b/IscDbc/Connection.h
@@ -524,7 +524,6 @@ public:
 	//virtual void		setSqlSubType ( short subtype ) = 0;
 	virtual void		setSqlLen ( short len ) = 0;
 	virtual short		getSqlMultiple () = 0;
-	virtual short		getSqlSubtype () = 0;
 	virtual short		getSqlLen () = 0;
 
 	// True for Firebird BINARY(n) / VARBINARY(n), i.e. CHAR(n) / VARCHAR(n)
--- a/IscDbc/Connection.h
+++ b/IscDbc/Connection.h
@@ -524,7 +524,6 @@ public:
 	//virtual void		setSqlSubType ( short subtype ) = 0;
 	virtual void		setSqlLen ( short len ) = 0;
 	virtual short		getSqlMultiple () = 0;
-	virtual short		getSqlLen () = 0;
 
 	// True for Firebird BINARY(n) / VARBINARY(n), i.e. CHAR(n) / VARCHAR(n)
 	// CHARACTER SET OCTETS at the wire layer (the FB4+ standard-compliant
--- a/IscDbc/Connection.h
+++ b/IscDbc/Connection.h
@@ -527,6 +527,12 @@ public:
 	virtual short		getSqlSubtype () = 0;
 	virtual short		getSqlLen () = 0;
 
+	// True for Firebird BINARY(n) / VARBINARY(n), i.e. CHAR(n) / VARCHAR(n)
+	// CHARACTER SET OCTETS at the wire layer (the FB4+ standard-compliant
+	// aliases).  Implementation in IscHeadSqlVar.h.
+	virtual bool		isBinary () = 0;
+	virtual bool		isVarBinary () = 0;
+
 	virtual char *		getSqlData() = 0;
 	virtual short *		getSqlInd() = 0;
 	virtual void		setSqlData( char *data ) = 0;
--- a/IscDbc/IscHeadSqlVar.h
+++ b/IscDbc/IscHeadSqlVar.h
@@ -100,6 +100,8 @@ public:
 	inline void	setSqlData ( char* data ) { sqlvar->sqldata = data; }
 
 	inline short	getSqlMultiple () { return sqlMultiple; }
+	inline short	getSqlSubtype () { return sqlvar->sqlsubtype; }
+	inline short	getSqlLen () { return sqlvar->sqllen; }
 	inline char *	getSqlData() { return sqlvar->sqldata; }
 	inline short *	getSqlInd() { return sqlvar->sqlind; }
 
--- a/IscDbc/IscHeadSqlVar.h
+++ b/IscDbc/IscHeadSqlVar.h
@@ -105,6 +105,28 @@ public:
 	inline char *	getSqlData() { return sqlvar->sqldata; }
 	inline short *	getSqlInd() { return sqlvar->sqlind; }
 
+	// Charset id for CHARACTER SET OCTETS.  In Firebird's XSQLVAR, when
+	// sqltype is SQL_TEXT or SQL_VARYING the sqlsubtype field carries the
+	// CHARACTER SET id (not a BLOB-style subtype); OCTETS is charset id 1
+	// (see IscDbc/MultibyteConvert.cpp CODE_CHARSETS(OCTETS, 1, 1)).
+	static constexpr short CHARSET_OCTETS = 1;
+
+	// True iff the slot describes a Firebird BINARY(n) — the FB4+ alias for
+	// CHAR(n) CHARACTER SET OCTETS.
+	inline bool		isBinary()
+	{
+		return sqlvar->sqltype == SQL_TEXT
+			&& sqlvar->sqlsubtype == CHARSET_OCTETS;
+	}
+
+	// True iff the slot describes a Firebird VARBINARY(n) — the FB4+ alias
+	// for VARCHAR(n) CHARACTER SET OCTETS.
+	inline bool		isVarBinary()
+	{
+		return sqlvar->sqltype == SQL_VARYING
+			&& sqlvar->sqlsubtype == CHARSET_OCTETS;
+	}
+
 	// not used
 	//void		setSqlInd( short *ind ) { sqlvar->sqlind = ind; }
 	//void		setSqlType ( short type ) { sqlvar->sqltype = type; }
--- a/IscDbc/IscHeadSqlVar.h
+++ b/IscDbc/IscHeadSqlVar.h
@@ -100,7 +100,6 @@ public:
 	inline void	setSqlData ( char* data ) { sqlvar->sqldata = data; }
 
 	inline short	getSqlMultiple () { return sqlMultiple; }
-	inline short	getSqlLen () { return sqlvar->sqllen; }
 	inline char *	getSqlData() { return sqlvar->sqldata; }
 	inline short *	getSqlInd() { return sqlvar->sqlind; }
 
--- a/IscDbc/IscHeadSqlVar.h
+++ b/IscDbc/IscHeadSqlVar.h
@@ -100,23 +100,25 @@ public:
 	inline void	setSqlData ( char* data ) { sqlvar->sqldata = data; }
 
 	inline short	getSqlMultiple () { return sqlMultiple; }
-	inline short	getSqlSubtype () { return sqlvar->sqlsubtype; }
 	inline short	getSqlLen () { return sqlvar->sqllen; }
 	inline char *	getSqlData() { return sqlvar->sqldata; }
 	inline short *	getSqlInd() { return sqlvar->sqlind; }
 
-	// Charset id for CHARACTER SET OCTETS.  In Firebird's XSQLVAR, when
-	// sqltype is SQL_TEXT or SQL_VARYING the sqlsubtype field carries the
-	// CHARACTER SET id (not a BLOB-style subtype); OCTETS is charset id 1
-	// (see IscDbc/MultibyteConvert.cpp CODE_CHARSETS(OCTETS, 1, 1)).
-	static constexpr short CHARSET_OCTETS = 1;
+	// Charset id for CHARACTER SET OCTETS, used to identify Firebird's
+	// FB4+ BINARY / VARBINARY types at the wire layer.  SqlProperties
+	// carries the charset id in a dedicated `sqlcharset` field (separate
+	// from `sqlsubtype`, which is documented as "BLOBs & Text types only";
+	// see IscDbc/Sqlda.h).  Charset id 1 = OCTETS per the driver's charset
+	// table at IscDbc/MultibyteConvert.cpp and per Firebird's
+	// RDB$CHARACTER_SETS system table.
+	static constexpr unsigned CHARSET_OCTETS = 1;
 
 	// True iff the slot describes a Firebird BINARY(n) — the FB4+ alias for
 	// CHAR(n) CHARACTER SET OCTETS.
 	inline bool		isBinary()
 	{
 		return sqlvar->sqltype == SQL_TEXT
-			&& sqlvar->sqlsubtype == CHARSET_OCTETS;
+			&& sqlvar->sqlcharset == CHARSET_OCTETS;
 	}
 
 	// True iff the slot describes a Firebird VARBINARY(n) — the FB4+ alias
@@ -124,7 +126,7 @@ public:
 	inline bool		isVarBinary()
 	{
 		return sqlvar->sqltype == SQL_VARYING
-			&& sqlvar->sqlsubtype == CHARSET_OCTETS;
+			&& sqlvar->sqlcharset == CHARSET_OCTETS;
 	}
 
 	// not used
--- a/OdbcConvert.cpp
+++ b/OdbcConvert.cpp
@@ -1661,25 +1661,60 @@ int OdbcConvert::notYetImplemented(DescRecord * from, DescRecord * to)
 
 int OdbcConvert::convGuidToString(DescRecord * from, DescRecord * to)
 {
-	char* pointer = (char*)getAdressBindDataTo((char*)to->dataPtr);
 	SQLLEN * indicatorTo = getAdressBindIndTo((char*)to->indicatorPtr);
 	SQLLEN * indicatorFrom = getAdressBindIndFrom((char*)from->indicatorPtr);
 
-	ODBCCONVERT_CHECKNULL( pointer );
+	if ( to->isIndicatorSqlDa )
+	{
+		ODBCCONVERT_CHECKNULL_SQLDA;
+	}
+	else
+	{
+		char* pointer = (char*)getAdressBindDataTo((char*)to->dataPtr);
+		ODBCCONVERT_CHECKNULL( pointer );
+	}
 
 	SQLGUID *g = (SQLGUID*)getAdressBindDataFrom((char*)from->dataPtr);
-	int len, outlen = to->length;
 
-	len = snprintf(pointer, outlen, "%08lX-%04X-%04X-%02X%02X-%02X%02X%02X%02X%02X%02X",
-		(unsigned int) g->Data1, g->Data2, g->Data3, g->Data4[0], g->Data4[1], g->Data4[2], g->Data4[3], g->Data4[4], g->Data4[5], g->Data4[6], g->Data4[7]);
+	// Format into a fixed 36+1 byte buffer first so the result is independent
+	// of to->length (which can be 0 for untyped VARCHAR parameters where
+	// Firebird has not assigned a precision yet).
+	char tmp[37];
+	int srcLen = snprintf(tmp, sizeof(tmp),
+		"%08X-%04X-%04X-%02X%02X-%02X%02X%02X%02X%02X%02X",
+		(unsigned int) g->Data1, g->Data2, g->Data3,
+		g->Data4[0], g->Data4[1], g->Data4[2], g->Data4[3],
+		g->Data4[4], g->Data4[5], g->Data4[6], g->Data4[7]);
+	if (srcLen < 0 || srcLen > 36) srcLen = 36;
 
-	if ( len == -1 ) len = outlen;
-
-	if ( to->isIndicatorSqlDa ) {
-		to->headSqlVarPtr->setSqlLen(len);
-	} else
-	if ( indicatorTo )
-		setIndicatorPtr(indicatorTo, len, to);
+	if ( to->isIndicatorSqlDa )
+	{
+		// Wire format: write into our own local buffer and redirect Firebird
+		// to read from there. This avoids buffer-size assumptions about the
+		// SQLDA-allocated wire buffer (which may be tiny — e.g., 0 or 2 bytes
+		// for an untyped `?` placeholder before its precision is known).
+		// The dispatch has already called setTypeText() so the wire is
+		// SQL_TEXT (no length prefix); we write exactly srcLen bytes and
+		// set sqllen accordingly.
+		if ( !to->isLocalDataPtr )
+			to->allocateLocalDataPtr( srcLen + 1 );
+		memcpy(to->localDataPtr, tmp, srcLen);
+		to->headSqlVarPtr->setSqlLen((short)srcLen);
+		to->headSqlVarPtr->setSqlData(to->localDataPtr);
+	}
+	else
+	{
+		char* pointer = (char*)getAdressBindDataTo((char*)to->dataPtr);
+		// Application output buffer: copy with NULL terminator.
+		int outlen = (int)to->length;
+		int copyLen = srcLen;
+		if (outlen > 0 && copyLen >= outlen) copyLen = outlen - 1;
+		if (copyLen < 0) copyLen = 0;
+		if (copyLen > 0) memcpy(pointer, tmp, copyLen);
+		if (outlen > 0) pointer[copyLen] = '\0';
+		if (indicatorTo)
+			setIndicatorPtr(indicatorTo, srcLen, to);
+	}
 
 	return SQL_SUCCESS;
 }
--- a/OdbcConvert.cpp
+++ b/OdbcConvert.cpp
@@ -1001,12 +1001,33 @@ ADRESS_FUNCTION OdbcConvert::getAdressFunction(DescRecord * from, DescRecord * t
 		}
 		break;
 	case SQL_C_GUID:
+		// On the wire side (to->isIndicatorSqlDa) Firebird describes a parameter
+		// slot as one of: BINARY(16) / CHAR(16) OCTETS (sqlsubtype == 1, sqllen == 16),
+		// CHAR/VARCHAR with a text charset, or genuine BINARY (FB4+). The canonical
+		// UUID is always ASCII, so a UTF8/wide-char wire still receives the same
+		// 36 narrow bytes — what differs is just whether we keep VARYING (length
+		// prefix) or convert to TEXT (no prefix).
+		if ( to->isIndicatorSqlDa )
+		{
+			if ( to->headSqlVarPtr->getSqlSubtype() == 1
+				&& to->headSqlVarPtr->getSqlLen() == 16 )
+				return &OdbcConvert::convGuidToBinary;
+			if ( to->conciseType == SQL_C_BINARY )
+				return &OdbcConvert::convGuidToBinary;
+			// Convert SQL_VARYING wire to SQL_TEXT so we can write 36 bytes
+			// directly without a length prefix; text subtype (charset) is
+			// preserved by setTypeText().
+			to->headSqlVarPtr->setTypeText();
+			return &OdbcConvert::convGuidToString;
+		}
 		switch(to->conciseType)
 		{
 		case SQL_C_CHAR:
 			return &OdbcConvert::convGuidToString;
 		case SQL_C_WCHAR:
 			return &OdbcConvert::convGuidToStringW;
+		case SQL_C_BINARY:
+			return &OdbcConvert::convGuidToBinary;
 		default:
 			return &OdbcConvert::notYetImplemented;
 		}
@@ -1678,6 +1699,44 @@ int OdbcConvert::convGuidToStringW(DescRecord * from, DescRecord * to)
 		(unsigned int) g->Data1, g->Data2, g->Data3, g->Data4[0], g->Data4[1], g->Data4[2], g->Data4[3], g->Data4[4], g->Data4[5], g->Data4[6], g->Data4[7]);
 
 	len = len == -1 ? outlen * sizeof( wchar_t ) : len * sizeof( wchar_t );
+
+	if ( to->isIndicatorSqlDa ) {
+		to->headSqlVarPtr->setSqlLen(len);
+	} else
+	if ( indicatorTo )
+		setIndicatorPtr(indicatorTo, len, to);
+
+	return SQL_SUCCESS;
+}
+
+// Write SQLGUID as 16 bytes in canonical UUID byte order (Data1/2/3 big-endian, Data4 unchanged).
+// Used when the target is BINARY(16) / CHAR(16) CHARACTER SET OCTETS or SQL_C_BINARY.
+int OdbcConvert::convGuidToBinary(DescRecord * from, DescRecord * to)
+{
+	char* pointer = (char*)getAdressBindDataTo((char*)to->dataPtr);
+	SQLLEN * indicatorTo = getAdressBindIndTo((char*)to->indicatorPtr);
+	SQLLEN * indicatorFrom = getAdressBindIndFrom((char*)from->indicatorPtr);
+
+	ODBCCONVERT_CHECKNULL( pointer );
+
+	SQLGUID *g = (SQLGUID*)getAdressBindDataFrom((char*)from->dataPtr);
+
+	unsigned char buf[16];
+	buf[0]  = (unsigned char)((g->Data1 >> 24) & 0xFF);
+	buf[1]  = (unsigned char)((g->Data1 >> 16) & 0xFF);
+	buf[2]  = (unsigned char)((g->Data1 >>  8) & 0xFF);
+	buf[3]  = (unsigned char)( g->Data1        & 0xFF);
+	buf[4]  = (unsigned char)((g->Data2 >>  8) & 0xFF);
+	buf[5]  = (unsigned char)( g->Data2        & 0xFF);
+	buf[6]  = (unsigned char)((g->Data3 >>  8) & 0xFF);
+	buf[7]  = (unsigned char)( g->Data3        & 0xFF);
+	memcpy(buf + 8, g->Data4, 8);
+
+	int outlen = (int)to->length;
+	int len = outlen < 16 ? outlen : 16;
+
+	if ( len > 0 )
+		memcpy(pointer, buf, len);
 
 	if ( to->isIndicatorSqlDa ) {
 		to->headSqlVarPtr->setSqlLen(len);
--- a/OdbcConvert.cpp
+++ b/OdbcConvert.cpp
@@ -1001,12 +1001,38 @@ ADRESS_FUNCTION OdbcConvert::getAdressFunction(DescRecord * from, DescRecord * t
 		}
 		break;
 	case SQL_C_GUID:
+		// Wire-side (input parameter binding) — Firebird's IPD describes the slot
+		// as either BINARY(16) / CHAR(16) OCTETS, a text VARCHAR/CHAR, or (FB4+)
+		// genuine BINARY. Route to dedicated wire-only conversions; leave the
+		// pre-existing convGuidToString / convGuidToStringW untouched for the
+		// app-side fetch path below.
+		if ( to->isIndicatorSqlDa )
+		{
+			// BINARY(16) / CHAR(16) CHARACTER SET OCTETS — 16 raw bytes
+			if ( to->headSqlVarPtr->getSqlSubtype() == 1
+				&& to->headSqlVarPtr->getSqlLen() == 16 )
+				return &OdbcConvert::convGuidToBinary;
+			// FB4+ BINARY/VARBINARY described directly as SQL_C_BINARY
+			if ( to->conciseType == SQL_C_BINARY )
+				return &OdbcConvert::convGuidToBinary;
+			// Text wire (VARCHAR/CHAR, any charset) — 36-char canonical UUID.
+			// setTypeText() converts SQL_VARYING to SQL_TEXT so convGuidToVarString
+			// can write the bytes without a length prefix.
+			to->headSqlVarPtr->setTypeText();
+			return &OdbcConvert::convGuidToVarString;
+		}
+		// App-side output (fetching a column described as SQL_GUID into the
+		// application's bound buffer). Currently unreachable because the column
+		// side of SQL_GUID mapping is task T5-5 (open), but kept intact for when
+		// that lands.
 		switch(to->conciseType)
 		{
 		case SQL_C_CHAR:
 			return &OdbcConvert::convGuidToString;
 		case SQL_C_WCHAR:
 			return &OdbcConvert::convGuidToStringW;
+		case SQL_C_BINARY:
+			return &OdbcConvert::convGuidToBinary;
 		default:
 			return &OdbcConvert::notYetImplemented;
 		}
@@ -1684,6 +1710,80 @@ int OdbcConvert::convGuidToStringW(DescRecord * from, DescRecord * to)
 	} else
 	if ( indicatorTo )
 		setIndicatorPtr(indicatorTo, len, to);
+
+	return SQL_SUCCESS;
+}
+
+// Write SQLGUID as 16 bytes in canonical UUID byte order (Data1/2/3 big-endian, Data4 unchanged).
+// Used when the target is BINARY(16) / CHAR(16) CHARACTER SET OCTETS or SQL_C_BINARY.
+int OdbcConvert::convGuidToBinary(DescRecord * from, DescRecord * to)
+{
+	char* pointer = (char*)getAdressBindDataTo((char*)to->dataPtr);
+	SQLLEN * indicatorTo = getAdressBindIndTo((char*)to->indicatorPtr);
+	SQLLEN * indicatorFrom = getAdressBindIndFrom((char*)from->indicatorPtr);
+
+	ODBCCONVERT_CHECKNULL( pointer );
+
+	SQLGUID *g = (SQLGUID*)getAdressBindDataFrom((char*)from->dataPtr);
+
+	unsigned char buf[16];
+	buf[0]  = (unsigned char)((g->Data1 >> 24) & 0xFF);
+	buf[1]  = (unsigned char)((g->Data1 >> 16) & 0xFF);
+	buf[2]  = (unsigned char)((g->Data1 >>  8) & 0xFF);
+	buf[3]  = (unsigned char)( g->Data1        & 0xFF);
+	buf[4]  = (unsigned char)((g->Data2 >>  8) & 0xFF);
+	buf[5]  = (unsigned char)( g->Data2        & 0xFF);
+	buf[6]  = (unsigned char)((g->Data3 >>  8) & 0xFF);
+	buf[7]  = (unsigned char)( g->Data3        & 0xFF);
+	memcpy(buf + 8, g->Data4, 8);
+
+	int outlen = (int)to->length;
+	int len = outlen < 16 ? outlen : 16;
+
+	if ( len > 0 )
+		memcpy(pointer, buf, len);
+
+	if ( to->isIndicatorSqlDa ) {
+		to->headSqlVarPtr->setSqlLen(len);
+	} else
+	if ( indicatorTo )
+		setIndicatorPtr(indicatorTo, len, to);
+
+	return SQL_SUCCESS;
+}
+
+// Write SQLGUID as the 36-char canonical UUID string into a Firebird text
+// parameter slot (VARCHAR/CHAR of any text charset). The wire-side SQLDA
+// buffer for an untyped `?` placeholder is sized for VARCHAR(0) — too small
+// for 36 bytes — so we stage the string in the DescRecord's local buffer
+// and redirect Firebird to read from there via setSqlData(), mirroring the
+// idiom transferStringToAllowedType already uses for app→wire string moves.
+// The dispatch in getAdressFunction has already called setTypeText() to
+// convert SQL_VARYING wires to SQL_TEXT (no length prefix), so we just
+// write 36 raw bytes and set sqllen.
+int OdbcConvert::convGuidToVarString(DescRecord * from, DescRecord * to)
+{
+	SQLLEN * indicatorFrom = getAdressBindIndFrom((char*)from->indicatorPtr);
+	SQLLEN * indicatorTo = getAdressBindIndTo((char*)to->indicatorPtr);
+
+	ODBCCONVERT_CHECKNULL_SQLDA;
+
+	SQLGUID *g = (SQLGUID*)getAdressBindDataFrom((char*)from->dataPtr);
+
+	char tmp[37];
+	int srcLen = snprintf(tmp, sizeof(tmp),
+		"%08X-%04X-%04X-%02X%02X-%02X%02X%02X%02X%02X%02X",
+		(unsigned int) g->Data1, g->Data2, g->Data3,
+		g->Data4[0], g->Data4[1], g->Data4[2], g->Data4[3],
+		g->Data4[4], g->Data4[5], g->Data4[6], g->Data4[7]);
+	if ( srcLen < 0 || srcLen > 36 ) srcLen = 36;
+
+	if ( !to->isLocalDataPtr )
+		to->allocateLocalDataPtr( srcLen + 1 );
+
+	memcpy(to->localDataPtr, tmp, srcLen);
+	to->headSqlVarPtr->setSqlLen((short)srcLen);
+	to->headSqlVarPtr->setSqlData(to->localDataPtr);
 
 	return SQL_SUCCESS;
 }
--- a/OdbcConvert.cpp
+++ b/OdbcConvert.cpp
@@ -1001,50 +1001,34 @@ ADRESS_FUNCTION OdbcConvert::getAdressFunction(DescRecord * from, DescRecord * t
 		}
 		break;
 	case SQL_C_GUID:
-		// Wire-side (input parameter binding): Firebird's IPD describes the
-		// slot as either BINARY(16) / VARBINARY(16) — i.e. (VAR)CHAR(16)
-		// CHARACTER SET OCTETS, the FB4+ standard-compliant aliases per the
-		// LangRef — or a text (VAR)CHAR. Route to dedicated wire-only
-		// conversions; leave the pre-existing convGuidToString /
-		// convGuidToStringW untouched for the app-side fetch path below.
-		if ( to->isIndicatorSqlDa )
+		switch(to->conciseType)
 		{
-			// BINARY(16) or VARBINARY(16). setTypeText() converts a varying
-			// OCTETS wire (VARBINARY) to fixed SQL_TEXT so convGuidToBinary
-			// can write at offset 0 with no VARYING length prefix; OCTETS
-			// charset is preserved.  For a fixed BINARY (SQL_TEXT already)
-			// it's a no-op.  Without this, Firebird reads the first GUID
-			// bytes as a VARYING length prefix and over-reads the buffer
-			// (SEGFAULT on the FB6 snapshot, silent on FB5).
-			if ( to->headSqlVarPtr->getSqlLen() == GUID_BINARY_LEN
-				&& ( to->headSqlVarPtr->isBinary()
-				  || to->headSqlVarPtr->isVarBinary() ) )
+		case SQL_C_CHAR:
+		case SQL_C_WCHAR:
+			if ( to->isIndicatorSqlDa )
 			{
+				// Input parameter.  Firebird describes BINARY(n) / VARBINARY(n)
+				// as (VAR)CHAR(n) CHARACTER SET OCTETS and the driver maps both
+				// to SQL_C_CHAR, so raw bytes vs. text is decided from the
+				// sqlvar, not from conciseType.  setTypeText() turns a
+				// SQL_VARYING slot into SQL_TEXT so the converters write at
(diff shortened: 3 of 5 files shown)


irodushka:
Honestly, I studied this PR and didn't understand a thing. Could you describe it more clearly—where the error was, how to reproduce it, and what the purpose of these changes was?..

irodushka:
I understand the code switching for string/binary, it's okay. But what's the purpose of total rewriting of the convGuidToString() foo? And why the convGuidToStringW() is left untouched?..

fdcastel:
Thanks for the review — fair feedback. I rewrote the commit so the diff is easier to read; here's the walkthrough.

**The bug.** An ODBC app calls `SQLBindParameter(SQL_C_GUID, SQL_GUID, …, ptr, 16, &len)` and hands the driver a 16-byte UUID. The driver returns `SQL_SUCCESS` but Firebird never sees the 16 UUID bytes:

- BINARY(16) / CHAR(16) CHARACTER SET OCTETS target: the bytes Firebird stores are `30 33 30 36 43 31 37 36 2D 45 34 30 31 2D 31 31` — the ASCII characters of the first 16 chars of the canonical UUID string. The driver was stringifying the GUID to 36 chars and then truncating to 16 to fit the column.
- VARCHAR target (e.g. `CHAR_TO_UUID(?)`): Firebird returns `expression evaluation not supported — Human readable UUID argument for CHAR_TO_UUID must have hex digit at position 2 instead of ""`. The driver was writing a wide-char (UTF-16) buffer into a narrow-char wire slot, so the embedded NULs made the string look empty.

Issue #295 has the full empirical trace including the two reproducers above (`SELECT UUID_TO_CHAR(?)` and `SELECT UUID_TO_CHAR(CHAR_TO_UUID(?))`). Real-world impact: duckdb/odbc-scanner binds DuckDB `::UUID` values exactly this way, so their Firebird UUID round-trip test (their [PR #169](https://github.com/duckdb/odbc-scanner/pull/169)) is parked on this.

**The fix shape, after the rewrite.** `convGuidToString` and `convGuidToStringW` are now **unchanged** — they remain the original short snprintf/swprintf helpers for the app-side fetch path (`SQLGetData(…, SQL_C_CHAR/WCHAR, …)` from a column described as SQL_GUID). That path is currently dead code because the column-side SQL_GUID mapping is open T5-5, but there's no reason to disturb them.

The wire-side fix lives in two new functions:

- `convGuidToBinary` — writes 16 bytes in canonical UUID byte order (Data1/2/3 swapped from x86 little-endian to big-endian, Data4 unchanged). Used when the wire is BINARY(16) / CHAR(16) OCTETS, or FB4+ BINARY/VARBINARY.
- `convGuidToVarString` — stages the 36-char canonical UUID in the DescRecord local buffer and points the wire's `sqldata` there via `setSqlData()`. This mirrors the idiom `transferStringToAllowedType` already uses, and sidesteps the fact that an untyped `?` placeholder is described by Firebird as VARCHAR(0) — its wire buffer only has room for the 2-byte length prefix, nowhere near enough for 36 chars.

The dispatch in `getAdressFunction` routes to the new functions on the wire side (`to->isIndicatorSqlDa`) and leaves the existing `SQL_C_CHAR / SQL_C_WCHAR` switch arms alone for the app-output direction.

`getSqlSubtype()` and `getSqlLen()` on `HeadSqlVar` are pure read-only accessors that let the dispatch distinguish `sqlsubtype == 1` + `sqllen == 16` (= BINARY/CHAR(16) OCTETS, must be raw bytes) from a regular text charset (= must be canonical UUID string).

History is squashed to one commit; force-pushed with `--force-with-lease`. Happy to split it back out if you'd rather see step-by-step commits.

fdcastel:
@irodushka Just a friendly ping on this PR when you have a chance. 🙂

This one (ideally together with #292 and #293) feels like a strong candidate for an rc2 release. That would let me finally close out the remaining DuckDB integration issue.

After that, we could get back on track with #287.

irodushka:
@fdcastel 

I'm lost in a tangle of different caliber troubles. Not related to FB or ODBC) Hardly doubt I'll be able to switch to those PRs this week, sorry(

fdcastel:
No worries. I know the feeling. 😅

Let me know when we can get back on track.

OdbcConvert.cpp
```diff
@@ -1001,12 +1001,38 @@ ADRESS_FUNCTION OdbcConvert::getAdressFunction(DescRecord * from, DescRecord * t
...
+				return &OdbcConvert::convGuidToBinary;
+			// FB4+ BINARY/VARBINARY described directly as SQL_C_BINARY
+			if ( to->conciseType == SQL_C_BINARY )
+				return &OdbcConvert::convGuidToBinary;
+			// Text wire (VARCHAR/CHAR, any charset) — 36-char canonical UUID.
+			// setTypeText() converts SQL_VARYING to SQL_TEXT so convGuidToVarString
+			// can write the bytes without a length prefix.
+			to->headSqlVarPtr->setTypeText();
+			return &OdbcConvert::convGuidToVarString;
+		}
```
irodushka, line 1032: maybe a good idea is to add `else` here, before the app-side logic switch...

OdbcConvert.cpp
```diff
@@ -1001,12 +1001,38 @@ ADRESS_FUNCTION OdbcConvert::getAdressFunction(DescRecord * from, DescRecord * t
...
 	case SQL_C_GUID:
+		// Wire-side (input parameter binding) — Firebird's IPD describes the slot
+		// as either BINARY(16) / CHAR(16) OCTETS, a text VARCHAR/CHAR, or (FB4+)
+		// genuine BINARY. Route to dedicated wire-only conversions; leave the
+		// pre-existing convGuidToString / convGuidToStringW untouched for the
+		// app-side fetch path below.
+		if ( to->isIndicatorSqlDa )
+		{
+			// BINARY(16) / CHAR(16) CHARACTER SET OCTETS — 16 raw bytes
+			if ( to->headSqlVarPtr->getSqlSubtype() == 1
```
irodushka: As for me, this condition looks indistinct. We gonna put data **from** app-side of the SQL_C_GUID type and **to**  smth with subtype == 1 and data length == 16... Something. And you suppose that that smth is BINARY/VARBINARY, for which FB 4+ uses combination of ~SUBTYPE==1 &&~ CHARACTER SET OCTETS && TYPE == (CHAR || VARCHAR)
But we do not check the wire-side to->conciseType == CHAR or VARCHAR. Maybe it makes sense to add such a cond?..

And I cannot find anything about subtype==1 for binary types... (see the comment below)

And maybe it's a good approarch to define a new inline foo headSqlVarPtr->isBinary() and headSqlVarPtr->isVarBinary()

irodushka: Please avoid using hardcoded `==1` or `==16`. It's much better to declare #define or static constexpr for compile-time consts.

OdbcConvert.cpp
```diff
@@ -1001,12 +1001,38 @@ ADRESS_FUNCTION OdbcConvert::getAdressFunction(DescRecord * from, DescRecord * t
...
+				&& to->headSqlVarPtr->getSqlLen() == 16 )
+				return &OdbcConvert::convGuidToBinary;
+			// FB4+ BINARY/VARBINARY described directly as SQL_C_BINARY
+			if ( to->conciseType == SQL_C_BINARY )
```
irodushka: hmmm, what's the difference from the condition above?.. Сan you please show me the place where to->conciseType is set to SQL_C_BINARY?...

That's what I see in FB4 docs:

- BINARY is a fixed-length data type, and is an SQL standard-compliant alias for **CHAR**(length) CHARACTER SET OCTETS
- VARBINARY is variable length binary type, and is an SQL standard-compliant alias for **VARCHAR**(length) CHARACTER SET OCTETS.

And I cannot find anything about subtype==1 for binary types... except AI advices) Perhaps the AI ​​got confused with Sybase RDB data types?.. AFAIC subtype 1 in FB means TEXT (for BLOBS at least). Can you show me some refs to FB documentation related to subtype == 1 for binary types?..

OdbcConvert.cpp
```diff
@@ -1688,6 +1714,80 @@ int OdbcConvert::convGuidToStringW(DescRecord * from, DescRecord * to)
...
+	buf[2]  = (unsigned char)((g->Data1 >>  8) & 0xFF);
+	buf[3]  = (unsigned char)( g->Data1        & 0xFF);
+	buf[4]  = (unsigned char)((g->Data2 >>  8) & 0xFF);
+	buf[5]  = (unsigned char)( g->Data2        & 0xFF);
+	buf[6]  = (unsigned char)((g->Data3 >>  8) & 0xFF);
+	buf[7]  = (unsigned char)( g->Data3        & 0xFF);
+	memcpy(buf + 8, g->Data4, 8);
+
+	int outlen = (int)to->length;
+	int len = outlen < 16 ? outlen : 16;
```
irodushka: If outlen < 16 then the truncation will occur. I don't think it's a SQL_SUCCESS... AFAIC there's a special return code for truncated data. And maybe it should be a plain error here because the truncated GUID becomes not unique and may lead to UB in DB model/logic.

OdbcConvert.cpp
```diff
@@ -1688,6 +1714,80 @@ int OdbcConvert::convGuidToStringW(DescRecord * from, DescRecord * to)
...
+int OdbcConvert::convGuidToVarString(DescRecord * from, DescRecord * to)
+{
+	SQLLEN * indicatorFrom = getAdressBindIndFrom((char*)from->indicatorPtr);
+	SQLLEN * indicatorTo = getAdressBindIndTo((char*)to->indicatorPtr);
+
+	ODBCCONVERT_CHECKNULL_SQLDA;
+
+	SQLGUID *g = (SQLGUID*)getAdressBindDataFrom((char*)from->dataPtr);
+
+	char tmp[37];
```
irodushka: Please do not hardcode.
```
static constexpr size_t GUID_CHAR_LEN = 36;
...
char tmp[GUID_CHAR_LEN + 1];
```

OdbcConvert.cpp
```diff
@@ -1688,6 +1714,80 @@ int OdbcConvert::convGuidToStringW(DescRecord * from, DescRecord * to)
...
+		"%08X-%04X-%04X-%02X%02X-%02X%02X%02X%02X%02X%02X",
+		(unsigned int) g->Data1, g->Data2, g->Data3,
+		g->Data4[0], g->Data4[1], g->Data4[2], g->Data4[3],
+		g->Data4[4], g->Data4[5], g->Data4[6], g->Data4[7]);
+	if ( srcLen < 0 || srcLen > 36 ) srcLen = 36;
+
+	if ( !to->isLocalDataPtr )
+		to->allocateLocalDataPtr( srcLen + 1 );
+
+	memcpy(to->localDataPtr, tmp, srcLen);
```
irodushka: Hmm and what if the `localDataPtr` size is < 36? Who may guarantee this?

OdbcConvert.cpp
```diff
@@ -1001,12 +1001,38 @@ ADRESS_FUNCTION OdbcConvert::getAdressFunction(DescRecord * from, DescRecord * t
...
+			if ( to->headSqlVarPtr->getSqlSubtype() == 1
+				&& to->headSqlVarPtr->getSqlLen() == 16 )
+				return &OdbcConvert::convGuidToBinary;
+			// FB4+ BINARY/VARBINARY described directly as SQL_C_BINARY
+			if ( to->conciseType == SQL_C_BINARY )
+				return &OdbcConvert::convGuidToBinary;
+			// Text wire (VARCHAR/CHAR, any charset) — 36-char canonical UUID.
+			// setTypeText() converts SQL_VARYING to SQL_TEXT so convGuidToVarString
+			// can write the bytes without a length prefix.
+			to->headSqlVarPtr->setTypeText();
```
irodushka, line 1030: Hmmm. Okay, let the type be text. But what about the destination buffer length? Look into my comment for convGuidToVarString() below.

irodushka:
@fdcastel found some time to throw some ~shi~ ideas at your PR)
Please cover the given code with test cases first and only then fix...
Cheers!

fdcastel:
Took the "cover the code with tests first" advice — and it immediately paid off.

- **Added `BindGuidToVarcharOctets16`** — binds `SQL_C_GUID` to `VARCHAR(16) CHARACTER SET OCTETS` (the wire form of native FB4+ `VARBINARY(16)`). The existing tests only covered the fixed `CHAR(16)` / `SQL_TEXT` slot; this is the `SQL_VARYING` binary wire, which was untested.
- **That test exposed a real bug, now fixed.** On the varying wire, `convGuidToBinary` wrote the 16 bytes at offset 0 with no 2-byte VARYING length prefix; Firebird then read the first GUID bytes as a length and over-read the buffer — a hard **SEGFAULT on the FB6 snapshot** (silently tolerated on FB5). Fix mirrors the text branch: the dispatch now calls `setTypeText()` for the OCTETS subtype-1/len-16 case so the varying wire becomes fixed `SQL_TEXT` and the offset-0 write is correct (no-op for an already-fixed `CHAR(16)` wire).
- **Dropped two dead `SQL_C_BINARY` dispatch arms.** Native BINARY/VARBINARY reach the dispatch as OCTETS subtype-1/len-16 (handled above), so the wire-side `conciseType == SQL_C_BINARY` arm only fired for a binary BLOB; the app-side arm is unreachable until column-side SQL_GUID mapping (T5-5 / #287) lands.

Rebased on current `master`. CI is green across the matrix. The new VARBINARY test is `SKIP_ON_FIREBIRD6`-guarded: even with the prefix fix, the FB6 snapshot still aborts this parameterized OCTETS-varying insert with a server-side "Stack overflow" — the same parameterized-query incompatibility already guarded across the suite — so it runs on FB 3/4/5 and skips on FB6.

Whenever you have a moment, a review would be much appreciated 🙏

irodushka:
Hi @fdcastel 

Okay, but the most of my notes (7 to be exact) are still pending...

fdcastel:
> Okay, but the most of my notes (7 to be exact) are still pending...

?

Sorry... What notes?

irodushka:
These)) didn't commit the review, sorry

fdcastel:
Pushed [`1571a97`](https://github.com/fdcastel/firebird-odbc-driver/pull/296/commits/1571a97) addressing all 8 review notes — per-comment responses are on the inline threads below.

Verified locally: clean build (0 warnings), FB 5.0.3 → all 4 `GuidParamBindingTest` cases pass, FB6 snapshot → 3 pass + `BindGuidToVarcharOctets16` skipped per the documented FB6 parameterized-OCTETS-varying limitation (unchanged from `84babbf`), full local suite **225 passed, 0 failed**. CI green on the full matrix ([run 26532371395](https://github.com/fdcastel/firebird-odbc-driver/actions/runs/26532371395)).

OdbcConvert.cpp
fdcastel: Done in [`1571a97`](https://github.com/fdcastel/firebird-odbc-driver/pull/296/commits/1571a97) — explicit `else` added so the wire-vs-app split is obvious instead of relying on the reader noticing that the wire-side block always returns.

OdbcConvert.cpp
fdcastel: Done in [`1571a97`](https://github.com/fdcastel/firebird-odbc-driver/pull/296/commits/1571a97). New inline accessors on `HeadSqlVar` / `IscHeadSqlVar`:

```cpp
inline bool isBinary()    { return sqlvar->sqltype == SQL_TEXT    && sqlvar->sqlsubtype == CHARSET_OCTETS; }
inline bool isVarBinary() { return sqlvar->sqltype == SQL_VARYING && sqlvar->sqlsubtype == CHARSET_OCTETS; }
```

The `sqltype` check is exactly your CHAR/VARCHAR-at-the-wire check (SQL_TEXT for fixed, SQL_VARYING for varying). The dispatch now reads:

```cpp
if ( to->headSqlVarPtr->getSqlLen() == GUID_BINARY_LEN
    && ( to->headSqlVarPtr->isBinary() || to->headSqlVarPtr->isVarBinary() ) )
```

fdcastel: Done in [`1571a97`](https://github.com/fdcastel/firebird-odbc-driver/pull/296/commits/1571a97). Three new named constants: `IscHeadSqlVar::CHARSET_OCTETS = 1`, plus `GUID_BINARY_LEN = 16` and `GUID_STRING_LEN = 36` in [`OdbcConvert.h`](https://github.com/fdcastel/firebird-odbc-driver/blob/fix/issue-295-sql-c-guid-param-binding/OdbcConvert.h). The widespread `sqlsubtype == 1` sites elsewhere in `Sqlda.cpp` (TINYINT detection, etc.) are left out of scope of this PR.

OdbcConvert.cpp
fdcastel: Two parts.

**(a) The arm is already gone.** I dropped the wire-side `conciseType == SQL_C_BINARY` branch in [`84babbf`](https://github.com/fdcastel/firebird-odbc-driver/pull/296/commits/84babbf) (the amend right after your first round of comments) for exactly the reason you give. Looks like your review was pinned against the pre-amend diff so it didn't pick that up. To answer your question literally: the only path that yields `conciseType == SQL_C_BINARY` for an IPD record is a binary BLOB via `JDBC_LONGVARBINARY` (see [`OdbcDesc.cpp:1282-1284`](https://github.com/fdcastel/firebird-odbc-driver/blob/fix/issue-295-sql-c-guid-param-binding/OdbcDesc.cpp#L1282-L1284)) — not a sane GUID-bind target.

**(b) The `subtype == 1` thing isn't an AI hallucination.** The XSQLVAR `sqlsubtype` field has two distinct meanings depending on `sqltype`:
- For `SQL_BLOB`: a BLOB subtype (0=binary, 1=TEXT, …) per the FB LangRef BLOB section.
- For `SQL_TEXT` / `SQL_VARYING`: the **CHARACTER SET id**. OCTETS has id 1 — see [`IscDbc/MultibyteConvert.cpp:48-55`](https://github.com/fdcastel/firebird-odbc-driver/blob/fix/issue-295-sql-c-guid-param-binding/IscDbc/MultibyteConvert.cpp#L48-L55): `CODE_CHARSETS( OCTETS, 1, 1 )`.

So `sqltype == SQL_TEXT && sqlsubtype == 1` *is* `CHAR(n) CHARACTER SET OCTETS`, which by the FB4 LangRef is `BINARY(n)`. The new `isBinary()` / `isVarBinary()` helper docstrings cite both refs so the next reader doesn't have to dig.

OdbcConvert.cpp
fdcastel: Done in [`1571a97`](https://github.com/fdcastel/firebird-odbc-driver/pull/296/commits/1571a97) — plain error. The branch now posts ODBC SQLSTATE `22001` "String data, right truncation - GUID requires 16 bytes" via `parentStmt->postError(...)` and returns `SQL_ERROR`, matching the pattern `transferStringToAllowedType` uses for `01004`. The dispatcher already enforces `sqllen == GUID_BINARY_LEN` so this branch is unreachable in practice — kept as defensive code with a comment.

OdbcConvert.cpp
fdcastel: Done in [`1571a97`](https://github.com/fdcastel/firebird-odbc-driver/pull/296/commits/1571a97): `char tmp[GUID_STRING_LEN + 1]`, with `GUID_STRING_LEN = 36` defined in [`OdbcConvert.h`](https://github.com/fdcastel/firebird-odbc-driver/blob/fix/issue-295-sql-c-guid-param-binding/OdbcConvert.h) and used consistently for the snprintf cap, the clamp, and the `allocateLocalDataPtr` size.

OdbcConvert.cpp
fdcastel: Real bug — thanks for catching it. Fixed in [`1571a97`](https://github.com/fdcastel/firebird-odbc-driver/pull/296/commits/1571a97).

The previous `if ( !to->isLocalDataPtr ) to->allocateLocalDataPtr( srcLen + 1 );` guard silently skipped (re)allocation when a smaller buffer already existed. Concrete failing case: an untyped `?` placeholder is described by Firebird as VARCHAR(0), so `DescRecord::allocateLocalDataPtr()` with no length argument defaults to `getBufferLength()` = 1 byte. If anything had allocated that 1-byte buffer earlier in the descriptor's life, the next call would have skipped reallocation and the 36-byte `memcpy` would have overflowed.

Fix: drop the guard, call `to->allocateLocalDataPtr( GUID_STRING_LEN + 1 )` unconditionally. `allocateLocalDataPtr()` frees the existing buffer first (see [`DescRecord.cpp:121-130`](https://github.com/fdcastel/firebird-odbc-driver/blob/fix/issue-295-sql-c-guid-param-binding/DescRecord.cpp#L121-L130)) so it's safe to call repeatedly, and it always yields exactly a 37-byte buffer.

OdbcConvert.cpp
fdcastel: Two layers, both fine after [`1571a97`](https://github.com/fdcastel/firebird-odbc-driver/pull/296/commits/1571a97):

1. **FB-side exec buffer.** `setTypeText()` mutates the sqlvar so `Sqlda::isExternalOverriden()` returns true on the next execute and `Sqlda::checkAndRebuild()` ([`IscDbc/Sqlda.cpp:397-475`](https://github.com/fdcastel/firebird-odbc-driver/blob/fix/issue-295-sql-c-guid-param-binding/IscDbc/Sqlda.cpp#L397-L475)) builds fresh `execMeta`/`execBuffer` sized for the new (TEXT) metadata.
2. **Driver-side staging buffer.** Independent of FB's rebuild, `convGuidToVarString` redirects the wire's `sqldata` via `setSqlData()` to our `localDataPtr` — which is now (with the fix for your previous note) guaranteed to be 37 bytes. So FB's rebuilt exec buffer is bypassed for the data payload entirely. Same idiom that `transferStringToAllowedType` uses for app→wire string moves.

irodushka:
Aha... Now I got it - there was no separate field for charset in the ISC API, so subtype was used.
But

```
    inline void bindProperties( Firebird::ThrowStatusWrapper& status, Firebird::IMessageMetadata* _meta, unsigned _index ) {
		index = _index;
		//
		offsetData = _meta->getOffset    ( &status, index );
		offsetNull = _meta->getNullOffset( &status, index );
		isNullable = _meta->isNullable   ( &status, index );
		sqlname    = _meta->getField     ( &status, index );
		relname    = _meta->getRelation  ( &status, index );
		aliasname  = _meta->getAlias     ( &status, index );
		//attention - ooapi may return type+bit if getType() is called from IMetadataBuilder!
		sqltype    = _meta->getType      ( &status, index ) & ~1;
		//
		sqlscale   = _meta->getScale     ( &status, index );
		sqlsubtype = _meta->getSubType   ( &status, index );
		sqlcharset = _meta->getCharSet   ( &status, index );
		sqllen     = _meta->getLength    ( &status, index );

		//OOAPI provides subtype & charset separately
		if (sqltype == SQL_TEXT || sqltype == SQL_VARYING)
		{
			sqlsubtype = sqlcharset;
		}
		//
		++index; //to make it 1-based)

		orgSqlProperties = *this; // save original props
		lastSqlProperties = *this;
	}
```	

and I think it's much more clear to specify `sqlvar->sqlcharset == CHARSET_OCTETS;` in is(Var)Binary foos.

IscDbc/Connection.h
```diff
@@ -524,6 +524,14 @@ class HeadSqlVar
 	//virtual void		setSqlSubType ( short subtype ) = 0;
 	virtual void		setSqlLen ( short len ) = 0;
 	virtual short		getSqlMultiple () = 0;
+	virtual short		getSqlSubtype () = 0;
```
irodushka: Obsolete now

IscDbc/IscHeadSqlVar.h
```diff
@@ -100,9 +100,33 @@ class IscHeadSqlVar : public HeadSqlVar
 	inline void	setSqlData ( char* data ) { sqlvar->sqldata = data; }
 
 	inline short	getSqlMultiple () { return sqlMultiple; }
+	inline short	getSqlSubtype () { return sqlvar->sqlsubtype; }
```
irodushka: already outdated

OdbcConvert.h
```diff
@@ -48,6 +48,13 @@ class DescRecord;
 class OdbcConvert;
 class OdbcStatement;
 
+// Canonical SQLGUID wire sizes.
+// - GUID_BINARY_LEN: 16 raw bytes, the BINARY(16) / VARBINARY(16) wire form.
+// - GUID_STRING_LEN: 36 chars (8-4-4-4-12 hex+dashes, no NUL terminator),
+//   the canonical text form written into VARCHAR/CHAR slots.
+static constexpr int    GUID_BINARY_LEN = 16;
```
irodushka: Both `int` or both `size_t`, why they're of different types? Looks inaccurate.

OdbcConvert.cpp
```diff
@@ -1001,14 +1001,50 @@ ADRESS_FUNCTION OdbcConvert::getAdressFunction(DescRecord * from, DescRecord * t
...
+			{
+				to->headSqlVarPtr->setTypeText();
+				return &OdbcConvert::convGuidToBinary;
+			}
+			// Text wire (VARCHAR/CHAR, any non-OCTETS charset) — 36-char
+			// canonical UUID.  setTypeText() converts SQL_VARYING to
+			// SQL_TEXT so convGuidToVarString can write the bytes without
+			// a length prefix.
+			to->headSqlVarPtr->setTypeText();
+			return &OdbcConvert::convGuidToVarString;
```
irodushka: I DONT understand the fundamental difference between convGuidToVarString (new) and convGuidToString (old). And I don't understand why convGuidTo**Var**String if we set sqltype to `TEXT` just before call.

IscDbc/Connection.h
fdcastel: Done in [`ccbcf93`](https://github.com/fdcastel/firebird-odbc-driver/pull/296/commits/ccbcf93) — dropped. After the `isBinary()` / `isVarBinary()` refactor in `1571a97`, `grep -n getSqlSubtype` returned only the declaration here and the inline impl in `IscHeadSqlVar.h`; no callers anywhere. Both gone now.

IscDbc/IscHeadSqlVar.h
fdcastel: Done in [`ccbcf93`](https://github.com/fdcastel/firebird-odbc-driver/pull/296/commits/ccbcf93) — dropped along with the matching pure-virtual decl in `Connection.h`.

OdbcConvert.h
fdcastel: Done in [`ccbcf93`](https://github.com/fdcastel/firebird-odbc-driver/pull/296/commits/ccbcf93) — unified to `int` for both:

```cpp
static constexpr int GUID_BINARY_LEN = 16;
static constexpr int GUID_STRING_LEN = 36;
```

Matches the `short → int` promotion at `getSqlLen() == GUID_BINARY_LEN` and lets the `(int)` casts inside `convGuidToWireString` (formerly `convGuidToVarString`) drop out.

OdbcConvert.cpp
fdcastel: Two parts.

**(a) The fundamental difference.** They write to opposite directions:

- `convGuidToString` (the pre-existing one) is the **app-side** writer — used in the `SQLGetData(..., SQL_C_CHAR, ...)` fetch path from a column described as `SQL_GUID`. It writes straight into the application's `to->dataPtr` and assumes the app has sized that buffer. That path is currently dormant; it'll wake up when the column-side `SQL_GUID` mapping lands with T5-5 / #287.
- `convGuidToVarString` / now-`convGuidToWireString` is the **wire-side** writer — what this PR adds, for `SQLBindParameter(SQL_C_GUID, ...)` against a text VARCHAR/CHAR parameter. The wire `sqldata` buffer is sized by Firebird from the IPD, and for an untyped `?` placeholder Firebird describes the slot as `VARCHAR(0)` — so the backing buffer is just 1 byte, nowhere near the 36 we need. Writing into it directly would corrupt the prefix and the surrounding wire buffer.

So the wire-side path stages the 36-char canonical UUID in `DescRecord::localDataPtr` (sized explicitly to `GUID_STRING_LEN + 1`) and redirects Firebird to read from there via `HeadSqlVar::setSqlData()` — the same idiom `transferStringToAllowedType` already uses for app→wire string moves. That's why the function exists separately.

**(b) The "Var" misnomer.** Fair catch. "Var" suggested `SQL_VARYING`, but the function actually runs against a `SQL_TEXT` wire (the dispatch calls `setTypeText()` right before). Renamed to **`convGuidToWireString`** in [`ccbcf93`](https://github.com/fdcastel/firebird-odbc-driver/pull/296/commits/ccbcf93) — and rewrote the leading docstring to spell out (a) above so the next reader doesn't have to ask.

fdcastel:
Done in [`ccbcf93`](https://github.com/fdcastel/firebird-odbc-driver/pull/296/commits/ccbcf93). Switched `isBinary()` / `isVarBinary()` to read `sqlvar->sqlcharset == CHARSET_OCTETS` instead of `sqlvar->sqlsubtype`. You're right — `SqlProperties` ([`IscDbc/Sqlda.h:34-49`](https://github.com/fdcastel/firebird-odbc-driver/blob/fix/issue-295-sql-c-guid-param-binding/IscDbc/Sqlda.h#L34-L49)) carries the charset id in a dedicated field, the comment on `sqlsubtype` literally says "datatype subtype - BLOBs & Text types only", and `CAttrSqlVar::bindProperties` ([`IscDbc/Sqlda.h:115-118`](https://github.com/fdcastel/firebird-odbc-driver/blob/fix/issue-295-sql-c-guid-param-binding/IscDbc/Sqlda.h#L115-L118)) only overlays `sqlcharset` into `sqlsubtype` for downstream compatibility. Reading the dedicated `sqlcharset` field is the semantically right one and removes the "two meanings of `sqlsubtype`" mental gymnastics entirely. `CHARSET_OCTETS` is now `unsigned` to match `sqlcharset`'s declared type, and the helper docstring is rewritten accordingly. Thanks for catching it.

irodushka:
Hi @fdcastel 

about https://github.com/FirebirdSQL/firebird-odbc-driver/pull/296#discussion_r3319046824

All these conv mechanix (OdbcConvert.cpp) is not mine, but I think I understand the initial idea. The switching is going primarily on `from` type and then - on `to` type. And it's symmetric and have no need to know the phy direction wire->app or visa versa. All the logic operates with the abstract `from` and `to` objs - I mean the `(DescRecord * from, DescRecord * to)`  basic contract. Well, if you need you can obtain the phy direction via the `bool DescRecord::isIndicatorSqlDa` property, but it may be used for exclusions only.

This approach, primarily fine and solid, subsequently was repeatedly distorted by countless updates & fixes)
But that's just the tip of the iceberg comparing to AI)))) [we Russians say - that's just the flowers, wait for the berries)]

I must think about it))

fdcastel:
Agreed. That’s the core issue. You're right that the skeleton is "from-then-to": the outer `switch (from->conciseType)` with an inner `switch (to->conciseType)` per arm.

On the symmetry point, I leaned on the existing pattern: `to->isIndicatorSqlDa` is already a primary dispatch axis throughout this function, not just an exclusion. A couple of examples in the current code:

- The NUMERIC arm picks the function pointer by direction, per `to` type — `OdbcConvert.cpp:456-489`:
  ```cpp
  case SQL_C_SLONG:
      if ( to->isIndicatorSqlDa )
          return &OdbcConvert::convTagNumericToLong;   // wire side
      return &OdbcConvert::convNumericToLong;          // app side
  ```
  (same for TINYINT / SHORT / FLOAT / DOUBLE / BIGINT / NUMERIC right there, and the DATE / TIME / TIMESTAMP / BINARY / CHAR / WCHAR arms do the same kind of thing.)

- And the Text arm even hoists a direction check *above* its `to`-switch and calls `setTypeText()` before returning a wire-only helper — `OdbcConvert.cpp:723-726`:
  ```cpp
  if ( to->isIndicatorSqlDa && to->headSqlVarPtr->isReplaceForParamArray() )
  {
      to->headSqlVarPtr->setTypeText();
      return &OdbcConvert::transferArrayStringToAllowedType;
  }
  switch(to->conciseType) // Text
  ```
  which is structurally the same move the GUID arm makes.

So the wire-vs-app split (`convGuidToString` app-side / `convGuidToWireString` wire-side) mirrors the established `conv*` vs `transfer*` split: the `transfer*` / `setSqlData`-redirect helpers only ever exist on the `isIndicatorSqlDa` side and have no app-side branch. 

The reason the GUID wire routing can't be expressed as a pure `to->conciseType` table is that BINARY-vs-text is decided at runtime from the SQLDA state (`getSqlLen() == 16 && (isBinary() || isVarBinary())`), not from the concise type. And the app-side `convGuidToString` / `convGuidToStringW` branch is dormant for now: it only fires once the column-side `SQL_GUID` mapping lands (T5-5 / #287).

Where I think your instinct is dead on: the GUID arm is the one place that puts `if (to->isIndicatorSqlDa)` *outermost* with the `to`-switch nested in the `else`, whereas most arms switch on `to` first and check direction per-case. That's a real ordering inconsistency, purely cosmetic. I'm happy to flip it to `to`-first to match the surrounding arms — it's behavior-identical — if you'd like that for consistency. Just say the word and I'll push it; otherwise I'll leave it as-is.

irodushka:
The worst thing that annoys me is the copypaste in convGuidToString/convGuidToWireString. These foos are 95% identical, except:
1. ODBCCONVERT_CHECKNULL... macro. IMHO it's not significant but it makes sense to deal with the macro zoo;
2. Direct printing to `to->dataPtr` buffer vs printing via stage tmp buffer & reallocating to->localBuffer;
3. Mention that the line `if ( len == -1 ) len = outlen;` in the legacy convGuidToString() foo is buggy. -1 means formatting error here == undefined data in the output buffer. len != GUID_STRING_LEN - always means some error.

Can you please explain me this:
```
// Wire-side counterpart of convGuidToString — writes SQLGUID as the 36-char
// canonical UUID into a Firebird text parameter slot (VARCHAR/CHAR of any
// non-OCTETS charset).  Where convGuidToString writes into an app-supplied
// buffer that the application sized (the SQLGetData fetch path, currently
// dormant until column-side SQL_GUID mapping lands in T5-5 / #287), this
// writes into the *wire* slot — and for an untyped `?` placeholder Firebird
// describes that slot as VARCHAR(0), whose default-sized backing buffer is
// just 1 byte, nowhere near the 36 we need.
```
?
It's not clear for me... As far I understand that's the case when we bind client-side SQLGUID **from** var (native 16-bytes GUID) to VARCHAR()/CHAR() **to** sqlda field, right? 

And the corresponding test is:
```
    SQLRETURN ret = SQLBindParameter(hStmt, 1, SQL_PARAM_INPUT,
        SQL_C_GUID, SQL_GUID, 16, 0, &guid, sizeof(guid), &guidInd);

    // Round-trip through CHAR_TO_UUID then UUID_TO_CHAR — exercises the
    // SQL_C_GUID → VARCHAR wire path on the way in.
    ret = SQLExecDirect(hStmt,
        (SQLCHAR*)"SELECT UUID_TO_CHAR(CHAR_TO_UUID(?)) FROM rdb$database",
        SQL_NTS);
```
?

I really wonder if this piece is correct.
The parameter is expected to be of CHAR/VARCHAR (TEXT) type and to have a size == 36. So why do you specify the dest type as SQL_GUID and the column size == 16?

Will it work if you write it like 
` SQLRETURN ret = SQLBindParameter(hStmt, 1, SQL_PARAM_INPUT, SQL_C_GUID, SQL_CHAR, 36, 0, &guid, sizeof(guid), &guidInd);` ?

Well, I can understand when you pass GUID in a text form (as CHAR(36)) to the parameter of SQL_GUID type. It's okay. But pass the 16-byte GUID directly to CHAR(36) column... Hmmm...

So as for me - you're trying to make a workaround to serve an incorrect parameter binding.
Look at this (the AI opinion:))) -
https://share.google/aimode/LHNJfo1dG0t7wgRcS

fdcastel:
Great catches! And one I want to (gently, with receipts 😄) push back on. Took them in order:

**1. The `len == -1` bug in `convGuidToString`: right on point.** 
- `snprintf` returning −1 is an encoding error, and `len = outlen` then hands back the whole buffer as if it were valid data. Worse than useless. 
- And your stronger framing is the correct invariant: a canonical GUID is *always* exactly 36 chars, so `len != 36` is by definition an error, never something to clamp and carry on. Fixed in `c8ae779` — non-36 is now a hard error (HY000), and a too-small destination raises `01004` with the full length reported, matching the ODBC "C to SQL: GUID" contract.

**2. The copy-paste: also right, dealt with.** 
- The genuinely duplicated part was the 11-arg format string living in two places. 
- Extracted a single `formatGuidCanonical()` helper that both functions now call; they keep only their distinct destination handling (app buffer vs staged `localDataPtr` + `setSqlData`). Same `c8ae779`.

**3. The macro zoo** (`ODBCCONVERT_CHECKNULL` vs `…_SQLDA`): agreed it's cosmetic; I'd just rather not fold it into this PR so the GUID change stays focused, but happy to do a small follow-up PR that tames the whole family at once after this one is solved.

**4. "Workaround to serve an incorrect parameter binding"**: this is the one where I think the binding is actually *correct*, and I went and gathered more evidence. Follow on:

4.a) First, your own question *"will it work if you write `SQLBindParameter(..., SQL_C_GUID, SQL_CHAR, 36, ...)`?"*: I tested it. On FB 5.0.3, round-tripping through `SELECT UUID_TO_CHAR(CHAR_TO_UUID(?))`:

```
C_GUID / SQL_GUID    / 16   ->  A0EEBC99-…-6BB9BD380A11   OK   (current — and what duckdb emits)
C_GUID / SQL_CHAR    / 36   ->  A0EEBC99-…-6BB9BD380A11   OK   (your suggestion)
C_GUID / SQL_VARCHAR / 36   ->  A0EEBC99-…-6BB9BD380A11   OK
C_GUID / SQL_GUID    / 0    ->  A0EEBC99-…-6BB9BD380A11   OK
```

So your binding works, but... all four land in the *same* `convGuidToWireString` 😅 
- For a prepared statement the driver converts to the parameter type Firebird described at prepare time (the `?` inside `CHAR_TO_UUID(?)` is text), not to the `ParameterType`/`ColumnSize` the app passed. 
- So `SQL_GUID, 16` vs `SQL_CHAR, 36` don't pick different code — the conversion is inherent to getting 16 bytes of GUID into a text parameter.

4.b) Second, the spec actually defines exactly this. The ODBC "C to SQL: GUID" conversion table ([learn.microsoft.com](https://learn.microsoft.com/en-us/sql/odbc/reference/appendixes/c-to-sql-guid)) lists `SQL_C_GUID` → `SQL_CHAR` (byte length ≥ 36, no error), `SQL_VARCHAR` (< 36 → `22001`), `SQL_WCHAR`/`SQL_WVARCHAR`, etc. So stringifying a bound GUID to its 36-char canonical form for a char/varchar parameter is the mandated behavior, not a hack. That same page even answers your "why `SQL_GUID, 16`?": *"The driver ignores the length/indicator value when converting data from the GUID C data type and assumes that the size of the data buffer is the size of the GUID C data type."* — i.e. `ColumnSize` is advisory for `SQL_C_GUID`, which is why the `16` (and the `0`, and the `36`) all behaved identically above.

4.c) Third, the real-world reason this exists: duckdb/odbc-scanner binds *exactly* `(SQL_C_GUID, SQL_GUID, 16)` ([uuid_type.cpp](https://github.com/duckdb/odbc-scanner/blob/main/src/types/uuid_type.cpp#L33-L44)). Before this fix the driver said `SQL_SUCCESS` and silently wrote garbage. So this isn't us inventing a use case — it's the binding real software emits, and the spec backs it.

Net: I **kept** the conversion (the spec sanctions it, your own proposed binding exercises the same path), and **folded in the bug fix + dedup you spotted**. Totally open to being wrong if you read that conversion table differently 😉. But I think on this one the berries turned out to be cherries. 🍒

fdcastel:
Sorry for the delay in finishing what you asked for; other professional tasks took my available time.

Pushed b303c08. What changed:

- `convGuidToWireString` is gone. `convGuidToString` is one function again; it branches on `to->isIndicatorSqlDa` only for the destination (Firebird side: staged in `localDataPtr` + `setSqlData`; application side: copy into the bound buffer with 01004), the same way the original function already branched for the length report.
- The `SQL_C_GUID` arm switches on `to->conciseType` first, like its neighbours. CHAR/WCHAR targets take the text or raw-bytes path; everything else falls to `notYetImplemented`, which restores the 07006 that master returned for a GUID bound to an INTEGER or DATE parameter.
- Any OCTETS slot receives the 16 raw bytes: 16 bytes or more stores the value, fewer is a 22001 from the driver. `getSqlLen()` is no longer needed and is removed.
- Both Firebird-side writers use `ODBCCONVERT_CHECKNULL_SQLDA`, so a value bound after a NULL clears the null flag. New tests: that case, 07006, the undersized and the oversized OCTETS slot.

The "VARCHAR(0) / 1-byte buffer" explanation I gave earlier was wrong, and the comments now say what actually happens:

- Reproducer B: with CHARSET=UTF8 Firebird describes the `?` in `CHAR_TO_UUID(?)` as CHAR(36) UTF8, 144 bytes. The driver maps UTF8 text to JDBC_WCHAR, so the old dispatch picked `convGuidToStringW` and wrote UTF-16 onto the wire; that is the "hex digit at position 2" error. With CHARSET=NONE the slot is CHAR(36) ASCII and the old code passed on Windows by accident, because `snprintf` is `_snprintf` there and 36 chars fit exactly with no terminator.
- Reproducer A: the old code stringified into the 16-byte OCTETS slot; `_snprintf` truncated and returned -1, and the `len == -1` line turned that into 16 bytes. So that line was MSVC truncation handling, not an encoding error as my last commit message said.
- The staging buffer is needed because the described slot can be VARYING, shorter than 36 bytes, or a wide-mapped text.

Verified locally on FB 5.0.3 with CHARSET=UTF8 and CHARSET=NONE: all 8 `GuidParamBindingTest` cases pass, full suite 229 passed, 0 failed under both charsets.

PR #298, opened 2026-05-11
Fix garbled diagnostic record on failed SQLDriverConnect

fdcastel:
## Summary

Every catch block that handled `std::exception` followed the same broken pattern:

```cpp
catch ( std::exception &ex ) {
    SQLException &exception = (SQLException&)ex;
    // virtual calls on `exception`...
}
```

The C-style cast is a `reinterpret_cast`. When the caught object was not in fact a `SQLException` (e.g. a `std::bad_alloc`, a raw `FbException` that bypassed the IscDbc rewrap, or anything else that propagated out of the IscDbc layer), the subsequent virtual dispatch through `exception` read through whatever bytes lay where `SQLException`'s vtable pointer would have been. On Linux that produced a diagnostic record with:

- empty `SQLSTATE`,
- a non-deterministic native code (varying every run),
- a one-byte message (`[`) whose reported length disagreed with its `strlen`.

ODBC clients (`unixODBC`'s `isql`, `pyodbc`, Tcl's `tdbc::odbc`, …) cannot recover useful information from such a record, so application-level retry/branch logic on `SQLSTATE` or native code is impossible.

## What this PR does

- Introduces an `ExceptionInfo` helper plus a new `postError(state, std::exception&)` overload in `IscDbc/SQLException.h` / `OdbcObject`. Both route through `dynamic_cast<SQLException*>`, falling back to `std::exception::what()` when the caught object isn't a `SQLException`.
- Converts all 64 occurrences of the broken pattern across `OdbcConnection`, `OdbcStatement`, `OdbcDesc`, `OdbcEnv`, and the Windows `OdbcJdbcSetup` tree to the safe path.
- Adds a NULL guard around `connection->close()` in `OdbcConnection::connect` so a throw inside `createConnection()` (before `connection` is assigned) no longer dereferences NULL during cleanup.
- Adds two regression tests in `tests/test_connect_options.cpp`:
  - `BadPasswordReturnsValidDiagRec` — verifies the diag record after a wrong-password failure (skips itself in embedded/trusted-auth setups that don't reject the password, including CI).
  - `NonexistentDatabaseReturnsValidDiagRec` — verifies the diag record after a connect to a nonexistent database path.

Both tests assert that after `SQLDriverConnect` fails the diag record has a 5-character `SQLSTATE`, a reported message length matching `strlen` of the message, and a message longer than a single byte.

## Verification

CI matrix (all green on the head of this PR):
- Windows x64 / x86 (Win32) / ARM64
- Linux x64 / ARM64
- Linux x64 with Valgrind

Tested against Firebird 5.0.3 and the Firebird master snapshot.

## Test plan
- [x] All existing tests still pass on the full Windows + Linux matrix
- [x] New `NonexistentDatabaseReturnsValidDiagRec` test passes on all platforms
- [x] New `BadPasswordReturnsValidDiagRec` test runs (skips gracefully when the server doesn't reject the bad password)
- [x] `grep -rn '(SQLException&)' *.cpp *.h IscDbc/ OdbcJdbcSetup/` returns no remaining unchecked casts

### DIFF
--- a/IscDbc/JString.cpp
+++ b/IscDbc/JString.cpp
@@ -199,7 +199,7 @@ const char* JString::getString()
 return (string) ? string : "";
 }
 
-JString::operator const char* ()
+JString::operator const char* () const
 {
 /**************************************
  *
--- a/IscDbc/JString.h
+++ b/IscDbc/JString.h
@@ -61,7 +61,7 @@ public:
 
 	void		Format (const char*, ...);
 	const char	*getString();
-	operator const char*();
+	operator const char*() const;
 	JString& operator = (const char *string);
 	JString& operator = (const JString& string);
 	JString& operator+=(const char *string);
--- a/IscDbc/SQLError.cpp
+++ b/IscDbc/SQLError.cpp
@@ -112,7 +112,7 @@ SQLError::~SQLError () throw()
 
 }
 
-int SQLError::getSqlcode ()
+int SQLError::getSqlcode () const
 {
 /**************************************
  *
@@ -128,7 +128,7 @@ int SQLError::getSqlcode ()
 return sqlcode;
 }
 
-int SQLError::getFbcode ()
+int SQLError::getFbcode () const
 {
 /**************************************
  *
@@ -144,7 +144,7 @@ int SQLError::getFbcode ()
 return fbcode;
 }
 
-const char *SQLError::getText ()
+const char *SQLError::getText () const
 {
 /**************************************
  *
@@ -160,7 +160,7 @@ const char *SQLError::getText ()
 return text;
 }
 
-SQLError::operator const char* ()
+SQLError::operator const char* () const
 {
 /**************************************
  *
@@ -193,7 +193,7 @@ SQLError::SQLError( int code, __int64 codefb, const char * txt, ...)
 	fbcode = (int) codefb;
 }
 
-const char* SQLError::getTrace()
+const char* SQLError::getTrace() const
 {
 	return stackTrace;
 }
--- a/IscDbc/SQLError.h
+++ b/IscDbc/SQLError.h
@@ -33,18 +33,18 @@ class SQLError : public SQLException
 public:
 	virtual int release();
 	virtual void addRef();
-	virtual const char* getTrace();
+	virtual const char* getTrace() const;
 	SQLError (int sqlcode, __int64 fbcode, const char *text, ...);
 	SQLError (SqlCode sqlcode, const char *text, ...);
 	SQLError (Stream *trace, SqlCode code, const char *txt,...);
 	~SQLError() throw();
 
-	virtual int			getFbcode ();
-	virtual int			getSqlcode ();
-	virtual const char	*getText();
+	virtual int			getFbcode () const;
+	virtual int			getSqlcode () const;
+	virtual const char	*getText() const;
 
 	//void		Delete();
-	operator	const char*();
+	operator	const char*() const;
 
 	int		fbcode;
 	int		sqlcode;
--- a/IscDbc/SQLException.h
+++ b/IscDbc/SQLException.h
@@ -42,10 +42,10 @@ class DllExport SQLException : public std::exception
 public:
 	//virtual void		addRef() = 0;
 	//virtual int		release() = 0;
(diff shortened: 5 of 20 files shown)


fdcastel:
@irodushka back from vacation? 😉

Another one similar to #296. As with that one, it would be great if this could make it into -rc2 as well.

This would allow us to run the [HammerDB benchmark](https://github.com/TPC-Council/HammerDB/discussions/881?utm_source=chatgpt.com) on Linux.

P.S. I’ve published a new unofficial build based on the latest `master` plus the 6 currently open PRs at https://github.com/fdcastel/firebird-odbc-driver/releases/tag/v3.5.1-rc1

irodushka:
Hi @fdcastel 

Yeah I'm online!
Trying to get back to the stream) Today is te very first day after the vacation, so don't expect too much))

What about this. AI - am I right about that?.. - is trying to solve a trivial bug (that's really _very_ bad, inaccurate, ugly, and the dev hands should be teared off!) with so comlpex & muddy means. A helper foo, and hell, dynamic_cast...

My Geniune Intelligence (I will abbreviate it to GI) tells me, that the much better approach is to separate exception handlers, in a standard way, like this:
```
try { ... }
catch (const SQLException& ex) {
    ...
}
catch (const std::exception& ex) {
    ...
}
```

If you're anxious about copypasting - you can use lambda or macrodef.

What about postError() - 1) I cannot understand what's the need for calling the extractExceptionInfo() helper in the overloaded version if you definitely know the exception type here 2) IMHO It's better to make a templated foo, and gate the logic with the `if constexpr (std::is_same_v<EX, std::exception>)` and so on

fdcastel:
Welcome back, man!

Sorry about this one. It was a quick fix I put together in a rush while trying to solve an issue on Linux. But ultimately I had to drop it (Unicode on Linux strikes again).

I’ll submit an update with your suggestions, and you can tell me what you think. But in the end, we can discard this one if you still don't like it (since I ended up not using it, after all).

fdcastel:
Thanks for the review @irodushka — fully agreed. Pushed [ec2e608](https://github.com/FirebirdSQL/firebird-odbc-driver/pull/298/commits/ec2e608) which scraps the `ExceptionInfo` helper / `dynamic_cast` and rewrites every catch as the standard two-handler form:

```cpp
catch (SQLException &ex)   { ... ex.getText() / ex.getSqlcode() ... }
catch (std::exception &ex) { ... ex.what() ... }
```

C++ catch matching picks the right handler statically; no runtime type check.

For `postError` I kept two simple, non-templated overloads — the existing `postError(state, SQLException&)` and a one-line `postError(state, std::exception&)` that just uses `ex.what()`. Each catch site dispatches to the right one from the static type of `ex`, so the new overload no longer does any cast.

`OdbcConnection::connect` keeps a small `rollbackPartialConnect` lambda to share the cleanup between its two catches (plus the `NULL` guard for `connection->close()` from the original PR).

Note on `const`: stayed with `catch (SQLException &ex)` (non-const) because `getText()` / `getSqlcode()` / `getFbcode()` are declared non-const virtuals in `IscDbc/SQLException.h`. Making them `const` would touch every IscDbc implementation — happy to do that in a follow-up if you'd prefer.

Full matrix is green on the new HEAD: Windows x64 / x86 / ARM64, Linux x64 / ARM64, Linux + Valgrind, against Firebird 5.0.3 and master.

irodushka:
@fdcastel 

Look at commit df437ea307d8f10a8e4eaae9ee80459657611103 (yes, I've just pushed a commit to your branch!) It's a simple way to make a const ref to our exceptions. I changed ref to const in the one place only (DsnDialog.cpp). If it works for you - please propagate it.

irodushka:
+++ 86c2d02330f63279e4e7444ed1833e94d768c265

fdcastel:
Done! Pulled both commits and ran a sweep over the remaining 63 catch pairs in [fd06bf2](https://github.com/FirebirdSQL/firebird-odbc-driver/pull/298/commits/fd06bf2). 

All `catch (SQLException &ex)` and `catch (std::exception &ex)` now read as `catch (const SQLException &ex)` / `catch (const std::exception &ex)` to match the style you set in `DsnDialog.cpp`.

